PR #3, opened 2018-06-08
Allow NumberOfResultsBefore or NumberOfResultsAfter to be 0

skinkie:
At this moment the user is forced to select at least one result after even if the user is only interested in one result before.
This commit allows to specify a 0 value, but also allow to omit the element, the system may infer the desired value.


sgrossberndt:
Fixes https://github.com/VDVde/OJP/issues/24

sgrossberndt:
Please amend the first commit so we have only one commit for this change in the history

sgrossberndt:
https://forum.vdv.de/viewtopic.php?f=88&t=157

skinkie:
@sgrossberndt could you tell me what I have to amend? You mean you want to 'rebase fast-forward'?

PR #4, opened 2018-06-08
Facilitate an extra Boarding and Alighting Time

skinkie:
The user might take extra boarding and alighting time towards and from a vehicle.
We may want to make this mode specific.


sgrossberndt:
As discussed in the meeting on 2018-11-16 this pull request is abandoned in favor of a new solution adding an `AdditionalTransferTime`of type `xs:duration` to `TripMobilityFilterGroup` in `TripParamStructure` and an `AdditionalTime` of type `xs:duration` to `IndividualTransportOptionsStructure`

PR #5, opened 2018-06-10
Fix siri_modes-v1.1 international

skinkie:


sgrossberndt:
This must be fixed upstream in SIRI, prepared in https://github.com/VDVde/SIRI/pull/2 to be probably included in 2.0q. We can then raise the included version of SIRI in OJP.

PR #6, opened 2018-06-10
Fix validation with xmllint

skinkie:
At this moment OJP based XML documents cannot be validated using xmllint.
This commit adds the required imports so the tool does function.


sgrossberndt:
This is a mix of fixing stuff in SIRI (2.0n did not include xml.xsd, but 2.0o does) and fixing how OJP includes SIRI.

sgrossberndt:
Closed as we want to use SIRI via git submodule

PR #7, opened 2018-08-09
TransferLimit should be a xs:nonNegativeInteger

skinkie:
Rationale a user might want to request a trip with 0 transfers.
And the TripStructure/Transfers is also a xs:nonNegativeInteger

Thanks @comlaterra for spotting this.


sgrossberndt:
Fixes https://github.com/VDVde/OJP/issues/28

PR #32, opened 2019-02-26
Allow multiple languages to be requested and returned.

skinkie:


sgrossberndt:
Not sure whether it was the right choice to merge this already - doesn't this need an updated upstream SIRI to work properly? Currently no time to look into it.

PR #35, opened 2019-07-08
Allow StopPoint information in LegIntermediate

skinkie:
Allow a reference to a `StopPlace` in a `StopPointGroup` instead of having it fixed as `StopPoint` by removing `StopPointName` and changing the reference from `StopPointRef` to `PlaceRef `.
This allows to show information from the `StopPointGroup` even if e.g. the definitive `StopPoint` assignment does not happen until a vehicle arrives at a platform.

### DIFF
--- a/OJP_LocationSupport.xsd
+++ b/OJP_LocationSupport.xsd
@@ -13,12 +13,7 @@
 			<xs:documentation>A stop point with id and name</xs:documentation>
 		</xs:annotation>
 		<xs:sequence>
-			<xs:element ref="siri:StopPointRef"/>
-			<xs:element name="StopPointName" type="InternationalTextStructure">
-				<xs:annotation>
-					<xs:documentation>Name or description of stop point for use in passenger information.</xs:documentation>
-				</xs:annotation>
-			</xs:element>
+			<xs:element ref="PlaceRefStructure"/>
 			<xs:element name="NameSuffix" type="InternationalTextStructure" minOccurs="0">
 				<xs:annotation>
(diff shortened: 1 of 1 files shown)


sgrossberndt:
Must be merged as `Squash and merge`.

sgrossberndt:
We followed the decision from the notes of the physical meeting on 16.11.2018 (where I was not available) - see https://forum.vdv.de/viewtopic.php?f=103&t=161#p616

I agree this circular reference is not a good choice. Maybe adding a new group allowing `StopPoint` and `StopPlace` might be a better option than changing the `StopPointGroup`.

pvgrumbkow:
I tried to remember the discussion in November. I cannot remember, whether we thought of the circular reference in PlaceStructure back than. 
StopPointGroup is used in 6 structures. 5 of them are in the context of service calls and there it seems to be OK to use PlaceRef. The last usage of StopPointGroup (in PlaceStructure via StopPointStructure) looks weird, if we approve this pull request.
Alternatives:
a) Approve this pull request and try to repair PlaceStructure afterwards.
b) Change the pull request in order to allow general places as service calls without changing the StopPointGroup. A new documentation would be needed then, too.
Opinions? a) is easier, but b) seems to be cleaner, right?

skinkie:
After a long discussion we have decided the following. We follow the Transmodel/NeTEx practise that for every stop there should be a ScheduledStopPoint. For unknown locations within a physical structure such as a StopPlace or a Stop where dynamic quay assignment happens a ScheduledStopPoint must also be introduced. This invalidates the pull request, considering we are modelling timed legs.

PR #36, opened 2019-07-08
Feature visually impaired

### LINKED ISSUE #26: 16. Visibility Impaired -  tactile pavement -> BaseTripContentFilter

skinkie:
Fixes #26

### DIFF
--- a/OJP_JourneySupport.xsd
+++ b/OJP_JourneySupport.xsd
@@ -561,6 +561,11 @@
 					<xs:documentation>Whether the result should include accessibility information.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element name="IncludeTactilePaving" type="xs:boolean" default="false" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Whether the result should include tactile paving guidance information to assist the visually impaired.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
 		</xs:sequence>
 	</xs:group>
 	<xs:annotation>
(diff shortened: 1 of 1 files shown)


OJP_JourneySupport.xsd
```diff
@@ -556,6 +561,11 @@
 					<xs:documentation>Whether the result should include accessibility information.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element name="IncludeTactilePaving" type="xs:boolean" default="false" minOccurs="0">
```
sgrossberndt: Why is an additional parameter `IncludeTactilePaving` a better choice than delivering the paving guidance information when `IncludeAccessibility` is `true`?

skinkie: Because someone that is in a Wheelchair literally hates a route with TactilePaving and prefers a route around it. While someone that is visibility impaired has great preference for it. Two different end users typically thrown in the same group of society having opposite needs.

pvgrumbkow: I agree to Stephan. A content filter "IncludeTactilePaving" is not needed here. The content filters define, which kind of information content is returned to the requestor. If he requests accessibility information, he will also get information on tactile pavement. The content filter has nothing to do with the actual routing. This would be influenced by the mobility filter "NoSight".
My conclusion: "NoSight" is sensible, "IncludeTactilePavement" is not.

skinkie:
@pvgrumbkow @sgrossberndt reverted the include tacticle paving.

skinkie:
Should we before merge also check the documentation?

Issue #50, opened 2019-11-04
Waiting Times


JonaVBB:
Extension for OJP v1.1

Add waiting times as a nonNegativeInteger for exchange points in ExchangePointResultStructure.

PR #58, opened 2020-01-29
Added WaitDuration to ExchangePointsResultStructure

pvgrumbkow:


pvgrumbkow:
Solves issue #50

OJP_Locations.xsd
```diff
@@ -343,6 +341,11 @@
 					<xs:documentation>Rough estimate of the travel duration from the specified location to this exchange point.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element name="WaitDuration" type="xs:duration" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Duration needed at this exchange point to change from one service to another. If a journey planning orchestrator puts together a trip at this exchange point, it hat to take care, that feeder and fetcher are at leat this duration apart. This is not true, if either feeding or fetching </xs:documentation>
```
sgrossberndt: "it ha_**s**_ to take care that feeder and fetcher are at lea_**s**_t this duration apart" instead of "it hat to take care, that feeder and fetcher are at leat this duration apart"
Is there text missing to complete the sentence after "This is not true, if either feeding or fetching "?

OJP_Locations.xsd
```diff
@@ -343,6 +341,11 @@
 					<xs:documentation>Rough estimate of the travel duration from the specified location to this exchange point.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element name="WaitDuration" type="xs:duration" minOccurs="0">
```
sgrossberndt: Would it make sense to add a `default="0"`?

OJP_Locations.xsd
pvgrumbkow: Sorry, yesterday I was so annoyed, that nothing worked, that I didn't check the annotations anymore.

pvgrumbkow: Solved in new commit.

OJP_Locations.xsd
pvgrumbkow: Added default PT0M into new commit.

ue71603:
We would prefer to use a structure like in NeTEx/Siri (e.g. PT3M). The current change also lacks in description. What is the waiting time? Minutes?

The type is siri/PositiveDurationType.

pvgrumbkow:
> We would prefer to use a structure like in NeTEx/Siri (e.g. PT3M). The current change also lacks in description. What is the waiting time? Minutes?
> 
> The type is siri/PositiveDurationType.
>
The type of the proposed field is xs:duration which has the same ISO format as PositiveDurationType of SIRI. So, it is not minutes, it is a duration.

Issue #51, opened 2019-11-04
Silent Fare Product


JonaVBB:
Extension for OJP v1.1

Add the possibility of a “silent” fare product. An additional boolean would define, that a specific fare product is not intended to be shown to the end user. Purpose would be to transport “incomplete” fare products, which have to be completed by other systems first.

PR #60, opened 2020-01-30
Added ProtoProduct into FareProductStructure

pvgrumbkow:
A proto product is a fare product, which shall not be shown to the user directly. Instead it carries fare product related information, which can be processed further. Can be used for tariffing in a distributed environment, where OJP services can deliver only parts of a fare product.


pvgrumbkow:
Solves issue #51

OJP_FareSupport.xsd
```diff
@@ -238,6 +237,11 @@
 				</xs:annotation>
 			</xs:element>
 			<xs:group ref="FareAuthorityGroup"/>
+			<xs:element name="ProtoProduct" type="xs:boolean" default="false" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Is the product a proto product? If yes, it should not be shown to the user directly. Instead a proto product transports fare product information, which usually has to be processed further. Default is false.</xs:documentation>
```
sgrossberndt: Extend by 
> Can be used for tariffing in a distributed environment, where OJP services can deliver only parts of a fare product.

?

pvgrumbkow: Shall I explain, why a proto fare product can be sensible in a distributed environment or shall we discard the "where ..." part of the sentence?

Issue #48, opened 2019-11-04
LegBoard/LegAlight


JonaVBB:
Extension for OJP v1.1

Define a field in LegBoard/LegAlight which consumes an arbitrary identifier of the service (usually not identical to the JourneyRef) at exactly that place (e.g. the train number). This shall be used for the search controller in EU-Spirit being able to recognize the usage of the same train by two adjacent systems. Alternatively, two such fields (for origin and destination) within TripStructure could be defined.

PR #61, opened 2020-01-30
Added LineId at boarding/alighting

pvgrumbkow:
Added LineIdAtBoarding in LegBoardStructure and LineIdAtAlighting in LegAlightStructure. This way, an arbitrary identifier for a service at boarding and alighting can be transported, which is independent from JourneyRef. This can be used in distributed environments in order to recognize, that two systems refer the same line/service while they still use their own internal references.

### DIFF
--- a/OJP_Trips.xsd
+++ b/OJP_Trips.xsd
@@ -547,9 +547,9 @@
 					</xs:sequence>
 				</xs:complexType>
 			</xs:element>
-			<xs:element name="LineIdAtBoarding" type="xs:normalizedString" minOccurs="0">
+			<xs:element name="DistributorInterchangeId" type="xs:normalizedString" minOccurs="0">
 				<xs:annotation>
-					<xs:documentation>Identifier of the boarded line/service at the boarding. This is not a reference. This identifier is used to recognize in a distributed environment, that two systems refer to the same line service (or service) while using their own internal references.</xs:documentation>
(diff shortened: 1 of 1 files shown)


pvgrumbkow:
Solves issue #48

OJP_Trips.xsd
```diff
@@ -548,6 +547,11 @@
 					</xs:sequence>
 				</xs:complexType>
 			</xs:element>
+			<xs:element name="LineIdAtBoarding" type="xs:normalizedString" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Identifier of the boarded line/service at the boarding. This is not a reference. This identifier is used to recognize in a distributed environment, that two systems refer to the same line service (or service) while using their own internal references.</xs:documentation>
```
sgrossberndt: > Identifier of the boarded line/service _during_ boarding. This is not a reference. This identifier is used to recognize _the line_ in a distributed environment _(two_ systems _referring_ to the same line service (or service) while using their own internal references).

With the current (slightly changed) documentation text it is still not clear to me what to expect in/from this element. Maybe enhance it a bit more?

What is meant by "line service (or service)"?

OJP_Trips.xsd
```diff
@@ -582,6 +586,11 @@
 					</xs:sequence>
 				</xs:complexType>
 			</xs:element>
+			<xs:element name="LineIdAtAlighting" type="xs:normalizedString" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Identifier of the alighted line/service at the alighting. This is not a reference. This identifier is used to recognize in a distributed environment, that two systems refer to the same line service (or service) while using their own internal references.</xs:documentation>
```
sgrossberndt: > Identifier of the alighted line/service _during_ alighting. This is not a reference. This identifier is used to recognize _the line_ in a distributed environment _(two_ systems _referring_ to the same line service (or service) while using their own internal references).

pvgrumbkow: Um, is your proposal to take over your edited text? "Line service (or service)" sounds really weird, simply "service" would be correct.

Issue #53, opened 2019-11-04
Flag for city names


JonaVBB:
Extension for OJP v1.1

Add a flag to PlaceStructure, whether the place is a city name result or not. Additionally it contains the system ID of the corresponding system, which is needed for subsequent requests after getting city name results from the RODI.

PR #62, opened 2020-01-30
Added ReferredSystemId in TopographicPlaceStructure

pvgrumbkow:
In an distributed environment, the process of place identification can be a two-steps process. In the first step, a topographic place (e.g. city, municipality) is identified from the user's input. In the second step, the system related to the topographic place is queried for places. In order to do so, the topographic places from the first step need to carry the information, to which system they relate.


pvgrumbkow:
Solves issue #53

PR #65, opened 2020-02-24
MinimizeContext

ue71603:
A flag ist added to the requests to minimize the context information that is returned.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+
+Changed_Files.zip
--- a/OJP_Trips.xsd
+++ b/OJP_Trips.xsd
@@ -168,6 +168,11 @@
 				<xs:annotation>
 					<xs:documentation>Whether the result should include operating day information - as encoded bit string and in natural language.</xs:documentation>
 				</xs:annotation>
+			</xs:element>	
+			<xs:element name="MinimizeResponseContext" type="xs:boolean" default="false" minOccurs="0">
+				<xs:annotation>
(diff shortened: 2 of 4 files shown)


pvgrumbkow:
What is "absolutely essential content"? It would be much better and clearer to add appropriate ContentFilters which cannot be misunderstood.

ue71603:
You are right. I will create a new pull request that is more in line with concrete ContentFilters.

.gitignore
```diff
@@ -0,0 +1,2 @@
+
+Changed_Files.zip
```
skinkie, line 2: This should be a separate pull request.

OJP_Trips.xsd
```diff
@@ -168,6 +168,11 @@
 				<xs:annotation>
 					<xs:documentation>Whether the result should include operating day information - as encoded bit string and in natural language.</xs:documentation>
 				</xs:annotation>
+			</xs:element>	
+			<xs:element name="MinimizeResponseContext" type="xs:boolean" default="false" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>hether the ResponseContext should be minimized to absolut essential content. As it is useful in many use cases. The default false means, that the context is sent.</xs:documentation>
```
skinkie, line 174: hether -> whether

pvgrumbkow:
This pull requests has two issues:
1. It's derived from master instead of changes_for_v1.1.
2. It should use specific content filters, e.g. "IncludePlaceContext", "IncludeSituationContext" as optional elements with default "true".

PR #68, opened 2020-03-17
add TravisCI and xmllint scripts for automated checks

haeckerbaer:
Add scripts for TravisCI functionality:
- check of xml syntax via xmllint
- possible checks of schema files via validation of standard examples (commented out because I don't have any OJP examples yet)
- TravisCI pushes back its changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: minimal
+
+before_script:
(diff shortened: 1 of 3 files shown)


skinkie:
Could someone verify that XML_IDENT is a tab?

PR #74, opened 2020-04-20
Add Travis integration

sgrossberndt:
* Improve xmllint
  * Change indent from one space to one tab
  * Enforce UTF-8 encoding
  * Remove XMLSpy comments
* Improve travis-commit
  * Allow Travis to push updates to all branches except master
  * Remove tagging

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: minimal
+
+before_script:
+  - sudo apt-get install -qq libxml2-utils
+
+script:
+  - bash .travis/xmllint-check.sh
+
+after_script:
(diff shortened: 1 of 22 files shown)


sgrossberndt:
> It is interesting to see that the style of each commit gets directly altered. Probably is good for the first issues.

I don't understand your comment. The current proposed solution is that Travis pushed a commit on each pull request and branch (except master) if linting yields differences to the original. Those can be squash-merged when applying to `changes_for_vx.x`. The first round to fix stuff is incorporated in this pull request (also in order to verify Travis does its job correctly).

PR #75, opened 2020-05-02
first load of test examples

ue71603:


ue71603:
The work was done by Daniel Rubli

skinkie:
1. We would like to have the examples tested with TravisCI.
2. Is there a reason why it was forked from master instead of the 1.1 branch?
3. Please join the next call and explain the background of these examples. It is appreciated.

ue71603:
I will create a new pull request based on 1.1.

PR #78, opened 2020-05-17
ExcludeContextSituations ExcludeContextPlaces

ue71603:
Two new parameters for excluding Context elements in the response.


ue71603:
used a completly wrong branch to do it....will do it again.

Issue #44, opened 2019-09-18
Rename LocationInformation to PlaceInformation


skinkie:

pvgrumbkow:
Handled by pull requests #82 and #99

PR #82, opened 2020-05-19
Replacement of Location with Place

pvgrumbkow:
According to TRANSMODEL a location is only a geographical position, while a place has a location along with other attributes. Within OJP the two terms were used inconsistently. This commit distinguishes the two terms properly. In most cases the term location was replaced by the term place. This applies to type and element names but also to annotations. This commit addresses issue #44.

### DIFF
--- a/OJP_JourneySupport.xsd
+++ b/OJP_JourneySupport.xsd
@@ -28,12 +28,12 @@
 		<xs:sequence>
 			<xs:element name="ViaPoint" type="PlaceRefStructure">
 				<xs:annotation>
-					<xs:documentation>Reference to specify the via location.</xs:documentation>
+					<xs:documentation>Reference to specify the via place.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
 			<xs:element name="DwellTime" type="xs:duration" minOccurs="0">
 				<xs:annotation>
-					<xs:documentation>Duration the passenger wants to stay at the via location. Default is 0.</xs:documentation>
+					<xs:documentation>Duration the passenger wants to stay at the via place. Default is 0.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
 		</xs:sequence>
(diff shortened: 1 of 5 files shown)


skinkie:
@pvgrumbkow is this something that is for 1.1 or for 2.0. Just asking, no judgement.

pvgrumbkow:
As far as I remember, this was planned for 1.1. This is a correction, not a functional extension.

OJP_JourneySupport.xsd
```diff
@@ -403,30 +403,30 @@
...
 					<xs:annotation>
-						<xs:documentation>Spatial location.</xs:documentation>
+						<xs:documentation>Static place.</xs:documentation>
 					</xs:annotation>
 				</xs:element>
-				<xs:element name="TripLocation" type="TripLocationStructure">
+				<xs:element name="TripPlace" type="TripPlaceStructure">
 					<xs:annotation>
-						<xs:documentation>Location within a (moving) vehicle.</xs:documentation>
+						<xs:documentation>Place within a (moving) vehicle.</xs:documentation>
```
skinkie, line 411: I really wonder if this is the wording used in Transmodel.

pvgrumbkow: In TRANSMODEL surely the term place is not used for what we call a TripPlace here. In lack of another term I simply used TripPlace. Perhaps we should request a term for this kind of "logical position within a vehicle" from TRANSMODEL. However, in the meantime TripLocation seems to be more wrong than TripPlace.

skinkie: Change the documentation part to something that does not suggest "seating".

pvgrumbkow: Please check again the new description.

herlitze (approved):

skinkie (approved):

Issue #71, opened 2020-04-02
Empty PublishedLineName


ue71603:
Hi everybody
In some cases there exists no LineName. However, the field is mandatory and can't be empty. Sending a blank does not seem a good idea.
Should we change the type? What are other opinions?
I think we need to coordinate this also with the SIRI CEN group. I ask there, too.
Kind regards
Matthias

`<xsd:complexType name="NaturalLanguageStringStructure">
<xsd:annotation>
<xsd:documentation>Tyoe for a string in a specified language.</xsd:documentation>
</xsd:annotation>
<xsd:simpleContent>
<xsd:extension base="PopulatedStringType">
<xsd:attribute ref="xml:lang" use="optional"/>
</xsd:extension>
</xsd:simpleContent>
</xsd:complexType>

Mir scheint also, als könnte man

<xsd:simpleType name="PopulatedStringType">
<xsd:annotation>
<xsd:documentation>A restriction of W3C XML Schema's string that requires at least one character of text.</xsd:documentation>
</xsd:annotation>
<xsd:restriction base="xsd:string">
<xsd:minLength value="1"/>
</xsd:restriction>
</xsd:simpleType>`

ue71603:
Or better, PublishedLineName is optional. That's the case in Siri. Then it is clear. Can we make it optional also in OJP?

pvgrumbkow:
The comment of PublishedLineName states: "Line name or service description as known to the public, f.e. "512", "S8" or "Circle Line".
In my point of view, this is always there. Every service has a line name or a service name, which is communicated to travelers. Therefore a traveler information API like OJP needs to have this kind of information as mandatory.
Could you provide examples, where a service does not have such a communicated name?

ue71603:
In Switzerland some regional trains don't have a line. Even the lines of IC1 and S1 are not really PublishedLineNames. The correct "published line" for both (and all six S1 that exist in Switzerland) is "1". The S1 in Lausanne, Bern, Basel, Zürich and the IC1 all share a pseudo line number of 1.

The RE Schaffhausen-Chur has no line. It only has a train number: Regional-Express 3472

This is by the way true for most train operators in Europe. They use train numbers.  Lines is mainly a concept of local transport. I know that we will get to lines also for trains, but currently to fill in the PublishedLineName is in general just not correct.

ue71603:
In the Swiss data and in the current restrictions, we will just but a "space" there. And I don't guess that's helpful.

[...]

PR #84, opened 2020-05-23
Optional PublishedLineName

ue71603:
In Siri PublishedLineName is optional and in NeTEx all possible equivalent elements in Line are optional as well and even the Ref to al Line in VehicleJourney is optional, so it should be optional here as well. This is alread

https://github.com/VDVde/OJP/issues/71

### DIFF
--- a/OJP_JourneySupport.xsd
+++ b/OJP_JourneySupport.xsd
@@ -58,7 +58,7 @@
 					<xs:documentation>[a specialisation of MODE in TMv6] an extended range of VEHICLE MODEs, aggregating them with some SUBMODEs</xs:documentation>
 				</xs:annotation>
 			</xs:element>
-			<xs:element name="PublishedLineName" type="InternationalTextStructure">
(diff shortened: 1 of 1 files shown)


pvgrumbkow:
Actually I don't see the necessity to follow Siri or NeTEx here. OJP is a service interface in order to give information to the traveler. Siri and NeTEx are data distribution interfaces/formats. 
Please have a look at the TimeLegStructure and its sub structures. If PublishedLineName was optional, there would be no mandatory element anymore, which gives the service of a TimedLeg a name. An information system accessible by OJP should be able to give a service an appropriate name to be given to the traveler. Otherwise we could end up with services without any description.
I disapprove this pull request.

ue71603:
Two things:
* When we name element the same in the Transmodel environment, they should have the same type. Otherwise use a different element name.
* After a longer discussion internally, we agree that we drop this pull request, However, it should be renamed to PublishedServiceName.

pvgrumbkow:
What is the type for the public service name in Transmodel?

PR #86, opened 2020-05-23
Relaxed requirement for id

ue71603:
Relaxed requirement for id. Mainly due to the definition for the German identifiers. But NeTEx also has this level of restriction.


pvgrumbkow:
This is a change within imported SIRI files. I think, this is not a good idea. This would have to be changed within SIRI.

PR #92, opened 2020-07-06
Service / mode extension

ue71603:
Service and ModeStructure didn't have the Extension element.

### DIFF
--- a/OJP_JourneySupport.xsd
+++ b/OJP_JourneySupport.xsd
@@ -356,6 +356,7 @@
 				</xs:annotation>
 			</xs:element>
 			<xs:element ref="SituationFullRef" minOccurs="0" maxOccurs="unbounded"/>
+			<xs:element name="Extension" type="xs:anyType" minOccurs="0"/>
 		</xs:sequence>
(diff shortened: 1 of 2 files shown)


herlitze (dismissed):

pvgrumbkow (dismissed):

pvgrumbkow:
After thinking again about the proposed change, I don't know anymore, whether it would be wise. The proposal is to add an "Extension" to the DatedJourneyGroup. This group is part of DatedJourneyStructure and of ContinousServiceStructure. In both cases the DatedJourneyGroup is a choice. The result is, that some DatedJourneys and some ContinousServices wonÄ't have an extension. Wouldn't it be better and perhaps also more intuitive, if DatedJourneyStructure and ContinousServiceStrucute would each get an "Extension"?
Please excuse, that I com with that one so late...

ue71603:
Update made according to Peter's suggestion.

pvgrumbkow (approved):

sgrossberndt (approved):

herlitze (approved):

PR #94, opened 2020-08-27
Rename PrivateCode to DomainCode

herlitze:


pvgrumbkow:
Thank you, Malte!

pvgrumbkow (approved):

sgrossberndt (approved):

skinkie:
I would not like this change. PrivateCode is a good term, but I think what we actually want is something that is more in the direction of a reference towards a StopPlace or Quay (with a global id). Maybe some people must convince me with an example that this is certainly not what we want here.

pvgrumbkow:
No consensus, therefore omitted.

PR #110, opened 2020-10-22
Moved ResponseContextStructure to OJP_RequestSupport.xsd

pvgrumbkow:
OJP_Places.xsd was not well-formed, because it used ResponseContextStructure without including it. Moved ResponseContextStructure from OJP_JourneySupport.xsd to OJP_RequestSupport.xsd, as it is a general structure and not only for journeys. Added includes of OJP_RequestSupport into several files.

### DIFF
--- a/OJP_JourneySupport.xsd
+++ b/OJP_JourneySupport.xsd
@@ -475,28 +475,6 @@
 			</xs:element>
 		</xs:sequence>
 	</xs:complexType>
-	<xs:complexType name="ResponseContextStructure">
-		<xs:annotation>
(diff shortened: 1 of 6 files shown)


pvgrumbkow:
Could you all please check, whether this branch validates in your frameworks?

ue71603 (approved):

sgrossberndt (approved):
I just cherry-picked these on top of https://github.com/VDVde/OJP/pull/115 and the validation of the examples does still work.
In general we should think whether these singular inclusions of xsds in others make sense or if we should add a `OJP_All.xsd` or the like which includes all sub-XSDs, but that is rather not the scope of this patch.

Issue #70, opened 2020-04-02
First batch of examples to test travis


ue71603:
The pull request must be adapted 
* for an identation with one tab (see https://gist.github.com/Protonk/3483490)
* The subdirectories for the xml-files need to be defined (currently subdirectory1 and subdirectory2 in examples
* The travis script works currently on the master branch. In SIRI we work on the integration branch, which is added to the master later. The group needs to define, what will be the workflow.
* Travis must be activated. Adrian tried to do it, but can't.  See https://travis-ci.com/. You go there and sign in with github. Then it should work.

`#!/bin/bash
/usr/bin/find . -name "*.xsd" -type f | while read i; do XMLLINT_INDENT=" " xmllint --pretty 1 "$i" > "$i.pretty"; mv "$i.pretty" "$i"; done; /usr/bin/find . -name "*.xml" -type f | while read i; do XMLLINT_INDENT=" " xmllint --pretty 1 "$i" > "$i.pretty"; mv "$i.pretty" "$i"; done; /usr/bin/find . -name "*.wsdl" -type f | while read i; do XMLLINT_INDENT=$"\t" xmllint --pretty 1 "$i" > "$i.pretty"; mv "$i.pretty" "$i"; done;
echo "finished formatting"
# xmllint --noout --schema OJP.xsd examples/subdirectory1/*xml examples/subdirectory2/*xml`

@haeckerbaer

ue71603:
For the activation @haeckerbaer said, tha one needs to be able to add travis in the organisation settings:
![image](https://user-images.githubusercontent.com/24227470/78237107-2bd10580-74db-11ea-805d-20c61021cd61.png)

sgrossberndt:
@haeckerbaer I have set up Travis and currently only allow Travis to push to the branch `travis_integration`
I have a successful build: https://travis-ci.com/github/VDVde/OJP/builds/160765382
but there are some problems:

- `travis-commit: failed to push git changes`
is shown in the build log (only visible when you expand the line 137)
Does Travis need additional permissions to be able to push?
- I would have expected the build to fail as there are formatting changes to the files which are not committed

Can you help here?

sgrossberndt:
We need to decide on a directory structure.
As we probably do not want to change the overall directory structure in v1.1 I propose to simply put them to `./examples`

ue71603:
Ok for me.

sgrossberndt:
Anyone willing to contribute some examples?

[...]

PR #115, opened 2020-11-11
Add examples and enable validation

sgrossberndt:
- Set OJP version to 1.1-dev
- Add examples and enable validation

Fixes: #70


OJP_JourneySupport.xsd
```diff
@@ -80,7 +77,7 @@
 					<xs:documentation>[a specialisation of MODE in TMv6] an extended range of VEHICLE MODEs, aggregating them with some SUBMODEs</xs:documentation>
 				</xs:annotation>
 			</xs:element>
-			<xs:element name="ProductCategory" type="ProductCategoryStructure">
+			<xs:element name="ProductCategory" type="ProductCategoryStructure" minOccurs="0">
```
sgrossberndt, line 80: Merge https://github.com/VDVde/OJP/pull/113 first

examples/switzerland/PlaceInformation_001_Stop_PlaceRef_Request.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<OJP xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.siri.org.uk/siri" xmlns:ojp="http://www.vdv.de/ojp" version="1.1-dev" xsi:schemaLocation="http://www.siri.org.uk/siri ../../OJP.xsd">
```
sgrossberndt, line 2: The definition `xsi:schemaLocation="http://www.siri.org.uk/siri ../../OJP.xsd"` is kind of a hack and up to now my biggest concern with the pull request.
SIRI is defined by `http://www.siri.org.uk/siri` and usually does not know anything about OJP. Only by referencing `../../OJP.xsd` here the root element definition `<OJP>` and the other elements from `OJP.xsd` are defined.

So OJP is based on SIRI but my XML/XSD knowledge is not good enough yet to propose a valid solution: Do we need an own namespace for this SIRI enhancement of `OJP.xsd` (apart from `xmlns:ojp="http://www.vdv.de/ojp"` which contains our own definitions not based on SIRI)?

Defining a `schemaLocation` is optional, also see https://stackoverflow.com/questions/4998063/one-xml-namespace-equals-one-and-only-one-schema-file:

> Some XML processors tolerate multiple <import> for the same namespace, and essentially amalgamate all of the schemaLocation into a single target.
> 
> Other processors are stricter, and decide that only one <import> per target namespace is valid. I think this is more correct, when you consider that schemaLocation is optional.
> 
> In addition to the VS and xmllint examples you gave, Xerces-J is also super-strict, and ignores subsequent <import> for the same target namespace, giving much the same error as xmllint does. XML Spy, on the other hand, is much more permissive (but then, XML Spy's validation is notoriously flaky)
> 
> To be safe, you should not have these multiple imports. A given namespace should have a single "master" document, which in turn has an <include> for each sub-document. This master is often highly artificial, acting only as a container. for these sub-documents.

examples/switzerland/PlaceInformation_001_Stop_PlaceRef_Request.xml
```diff
@@ -0,0 +1,23 @@
...
+	<OJPRequest>
+		<ServiceRequest>
+			<RequestTimestamp>2020-01-19T12:00:00Z</RequestTimestamp>
+			<RequestorRef>IRMA</RequestorRef>
+			<ojp:OJPPlaceInformationRequest>
```
sgrossberndt, line 7: I modified @ue71603 s examples from https://github.com/VDVde/OJP/pull/85/files to make them valid with 1.1-dev (`OJPPlaceInformationRequest` instead of `OJPLocationInformationRequest`, `ojp:Name` instead of `ojp:LocationName`, ...)

sgrossberndt (commented):
Most of my review comments are explanations, the biggest issue to discuss in my opinion is https://github.com/VDVde/OJP/pull/115/files#diff-0e9e7bd8d07a9a105056d18de9859da979d8cc797cdf56ef48f164970d1f4664R2

This branch is based on https://github.com/VDVde/OJP/pull/113 which should be merged first.

OJP_Requests.xsd
```diff
@@ -1,15 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<xs:schema xmlns="http://www.vdv.de/ojp" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:siri="http://www.siri.org.uk/siri" xmlns:xsd="http://www.w3.org/2001/XMLSchema" targetNamespace="http://www.vdv.de/ojp" elementFormDefault="qualified" attributeFormDefault="unqualified" id="OJP_siri_service">
-	<!-- ===Dependencies ======================================= -->
-	<xs:import namespace="http://www.siri.org.uk/siri" schemaLocation="./siri/siri_requests-v2.0.xsd"/>
-	<!-- ===OJP Services ======================================= -->
+<xs:schema xmlns="http://www.vdv.de/ojp" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:siri="http://www.siri.org.uk/siri" targetNamespace="http://www.vdv.de/ojp" elementFormDefault="qualified" attributeFormDefault="unqualified" id="OJP_siri_service">
```
sgrossberndt, line 2: `xmlns:xsd="http://www.w3.org/2001/XMLSchema"` was unused and has been removed

sgrossberndt:
Ready for review. As written before: https://github.com/VDVde/OJP/pull/115#discussion_r521215939 I am not sure if it is a valid behaviour to "redefine" the SIRI namespace like that..

pvgrumbkow (approved):

herlitze (approved):

OJP_Common.xsd
```diff
@@ -1,11 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <xs:schema xmlns="http://www.vdv.de/ojp" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:siri="http://www.siri.org.uk/siri" targetNamespace="http://www.vdv.de/ojp" elementFormDefault="qualified" attributeFormDefault="unqualified">
-	<!-- ===Dependencies ======================================= -->
-	<xs:import namespace="http://www.siri.org.uk/siri" schemaLocation="./siri_model/siri_reference-v2.0.xsd"/>
-	<xs:import namespace="http://www.siri.org.uk/siri" schemaLocation="./siri_model/siri_operator_support-v2.0.xsd"/>
-	<!-- ===========================================================================================================-->
+	<xs:import namespace="http://www.siri.org.uk/siri" schemaLocation="./siri_model/siri_all-v2.0.xsd"/>
 	<xs:include schemaLocation="OJP_ModesSupport.xsd"/>
 	<xs:include schemaLocation="OJP_FacilitySupport.xsd"/>
+	<xs:include schemaLocation="OJP_Requests.xsd"/>
```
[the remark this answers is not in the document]
sgrossberndt: Removed from this pull request and solved in another way in #117 by adding `OJP_All.xsd` in each file

PR #126, opened 2020-11-30
Examples switzerland 2

ue71603:
I updated all examples and checked them with xmllint.


sgrossberndt (changes requested):
Thanks for enhancing the examples, @ue71603 ! :)

I see some issues with the current state though:
We will have to adapt all examples when we are changing the schema.
All given examples should add value, differ and be as short as possible (yes, SER_02_response.xml, I am looking at you...)
20.000 lines of StopEvents are a bit too much in my opinion (and very repetitive too)

At least PIR_01_response.xml, PIR_02_request.xml, SER_01_response.xml and SER_02_request.xml are either missing or have wrong names.
The concept was meant as PIR_01_request.xml has its answer in PIR_01_response.xml

The names of the examples should reflect the differences (at least in my opinion) in order to pick the example you want easily - thats why I chose the naming scheme
PlaceInformation_001_Stop_PlaceRef_Request.xml 
PlaceInformation_001_Stop_PlaceRef_Response.xml 
PlaceInformation_002_Stop_InitialInput_Request.xml
PlaceInformation_002_Stop_InitialInput_Response.xml
Trip_001_Request.xml
Trip_001_Response.xml

ue71603:
I will adapt this before tomorrow. the important thing for me was that the xslt allows me to update 1.0--> 1.1 automatically. With the xslt I probably also can update all examples in one go, when I know what changed.

ue71603:
We could also switch to have tests: Meaning each change needs to have an example that covers it.

ue71603:
New names and shortened drastically.

In future I should add some more examples to test for specific new features.

skinkie:
You might want to check the "version" element, even if it is current a 'dev' version.

```
PIR_02_response_search_stops_by_name.xml:2: element OJP: Schemas validity error : Element '{http://www.siri.org.uk/siri}OJP', attribute 'version': The value '1.1-dev' does not match the fixed value constraint '1.0'.
```

ue71603:
will do.

ue71603:
included in new PR: https://github.com/VDVde/OJP/pull/334

PR #127, opened 2020-11-30
Added private modes

pvgrumbkow:
Added some private modes to IndividualModesEnumeration, e.g. scooter, ride-pool-car, car-sharing, cycle-sharing, scooter-sharing.

### DIFF
--- a/OJP/OJP_ModesSupport.xsd
+++ b/OJP/OJP_ModesSupport.xsd
@@ -17,6 +17,11 @@
 			<xs:enumeration value="others-drive-car"/>
 			<xs:enumeration value="motorcycle"/>
 			<xs:enumeration value="truck"/>
+			<xs:enumeration value="scooter"/>
+			<xs:enumeration value="ride-pool-car"/>
+			<xs:enumeration value="car-sharing"/>
+			<xs:enumeration value="cycle-sharing"/>
+			<xs:enumeration value="scooter-sharing"/>
 		</xs:restriction>
 	</xs:simpleType>
 	<xs:simpleType name="PrivateModesEnumeration">


sgrossberndt:
Do the names of those modes have to be aligned with some standard?
"cycle-sharing" vs "bicycle-sharing"

skinkie:
> Do the names of those modes have to be aligned with some standard?
> "cycle-sharing" vs "bicycle-sharing"

Let's ask the NeTEx new modes guys?

pvgrumbkow:
Actually, I did not have a look at NeTEx. the result of our discussion in the last call was, that this extension of the IndividualModesEnumeration should be quickly extended by usual modes. In parallel we planned to place an issue at SIRI for extending the PrivateModeChoiceGroup by standard individual modes like "walk", "own-bike", "own-car". If they do so, we could use this SIRI:PrivateModeChoiceGroup withint OJP 2.0.
However, I propose to extend the current enumeration for OJP 1.1 in a very simple way (keeping the current wording) and to plan a proper extension via SIRI for OJP 2.0.

ue71603:
Then we should add a note, that this may change in v2.0 when being more conform with Siri.

skinkie (approved):
Approve this with the condition that within OJP 2.0 this will solved by SIRI not by custom lists.

sgrossberndt (approved):

karsa-mistmere:
Can you please explain the motivation behind including sharing services within individual travel modes? (That is, if I understrand correctly what car-sharing, cycle-sharing and scooter-sharing mean).
Aren't sharing services supposed to be represented via a SharingService tag in ContinuousLeg trip legs?
Thank you for your reply in advance.

pvgrumbkow:
Yes, within responses sharing services are supposed to fill the SharingService element in ContinuousLeg. Still an explicit mode type does not conflict with this information.
But actually, the extension of IndividualModes was necessary for trip requests. Within requests it was beforehand not possible to allow sharing services. That was the reason to extend the IndividualModes.

PR #157, opened 2021-04-21
Removed annotations in choices

herlitze:
I did not find any annotations in xs:sequences.


pvgrumbkow (approved):

sgrossberndt (approved):
Removed annotation in OJP_Places.xsd could also be moved to line 16 (directly inside `PlaceInformationRequestGroup`) but I guess we'll rework /add sth there anyway when aligning with the Word documentation.

PR #174, opened 2021-11-28
Update siri_types-v2.0.xsd

phirzel:
the given path "../xml/xml.xsd" is not in the repository, assuming the fix as solution


sgrossberndt:
Thanks for the contribution. This problem has already been solved in the upcoming release v1.1 in branch [changes_for_v1.1](https://github.com/VDVde/OJP/tree/changes_for_v1.1) in commit https://github.com/VDVde/OJP/commit/9187b3d1d20753612d928b192ee6a021397dfb75#diff-ae7f8df42f6b8eadccdf20fee95f233b9b31a4d1c4806b52e212bcd5b90246dd which will soon be released.

PR #178, opened 2021-12-24
Quick fix that allows validation with xmllint

### LINKED ISSUE #177: Create a branch (1.0.2) that validates with xmllint
xmllint --noout --nowarning --schema OJP.xsd example.xml folgenden Fehler: failed to load external entity "xml/xml.xsd

I will then add a wiki page about that.

skinkie:
There are still redundant namespace imports.

Fix #177


1tchy:
I can approve, this fixes the validation issue for my case.

phirzel (commented):
do we need an "own" xml/xml.xsd file? see my proposition https://github.com/phirzel/OJP/commit/842a5e5e350eb5fcd60f709662243a614968cb45 (though I am irrelevant for this decision)

All other include adaptions look rather good to me -> I had problems with some of those with JAXB and OJP.xsd and hope this will fix buggy includes

skinkie:
Change with the xml.xsd from https://raw.githubusercontent.com/VDVde/OJP/changes_for_v1.1/siri/xml/xml.xsd

xml/xml.xsd
```diff
@@ -0,0 +1,160 @@
+<?xml version="1.0"?>
```
sgrossberndt, line 1: This change should use the more recent `xml.xsd` from https://raw.githubusercontent.com/VDVde/OJP/changes_for_v1.1/siri/xml/xml.xsd instead.

sgrossberndt (changes requested):

xml/xml.xsd
skinkie: This is fixed. Can you give your blessing?

sgrossberndt (approved):

ue71603 (approved):

PR #201, opened 2022-05-26
Add `PublicCode` of the journey to DatedJourney and ContinuousService

ue71603:
mostly as View to present things to customers


AndreasAtSBB (dismissed):
PublishedJourneyName added after JourneyRef
operatorName added after OperatorRef

OJP/OJP_JourneySupport.xsd
```diff
@@ -63,6 +63,11 @@
 		</xs:annotation>
 		<xs:sequence>
 			<xs:element ref="JourneyRef"/>
+			<xs:element name="PublishedJourneyName" type="xs:normalizedString" minOccurs="0">
```
Aurige, line 66: What do you mean bt "published" ? why not just "JourneyName" ?

ue71603: changed to PublicCode. Probably better as you suggested.

Aurige (dismissed):

OJP/OJP_JourneySupport.xsd
```diff
@@ -93,6 +98,11 @@
 				</xs:annotation>
 			</xs:element>
 			<xs:element ref="OperatorRef" minOccurs="0"/>
+			<xs:element name="OperatorName" type="xs:normalizedString" minOccurs="0">
```
sgrossberndt, line 101: Should this be of type `InternationalTextStructure`?

OJP/OJP_JourneySupport.xsd
```diff
@@ -63,6 +63,11 @@
 		</xs:annotation>
 		<xs:sequence>
 			<xs:element ref="JourneyRef"/>
+			<xs:element name="PublicCode" type="xs:string" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Descriptive public code of the SERVICE JOURNEY.</xs:documentation>
```
sgrossberndt, line 68: Could you please add an example?

OJP/OJP_JourneySupport.xsd
```diff
@@ -63,6 +63,11 @@
 		</xs:annotation>
 		<xs:sequence>
 			<xs:element ref="JourneyRef"/>
+			<xs:element name="PublicCode" type="xs:string" minOccurs="0">
```
sgrossberndt, line 66: Why is `OperatorName` below added as `type="xs:normalizedString` and `PublicCode` here as `type="xs:string"`? Wouldn't `type="xs:normalizedString` make sense here as well?

https://www.w3schools.com/xml/schema_dtypes_string.asp

OJP/OJP_JourneySupport.xsd
ue71603: updated

OJP/OJP_JourneySupport.xsd
ue71603: Not really. However, I harmonized with the NeTEx definition. Perhaps @Aurige has an example.

OJP/OJP_JourneySupport.xsd
ue71603: changed

Aurige (approved):

skinkie (approved):

sgrossberndt (changes requested):
I still do not understand what a "public code" of a journey is. Is it some kind of ID? Should it be shown to customers? Is it meant to refer to journeys by it? If yes, in which context? Is this defined anywhere so we could add this information to the annotation?

Aurige (dismissed):

Aurige:
> I still do not understand what a "public code" of a journey is. Is it some kind of ID? Should it be shown to customers? Is it meant to refer to journeys by it? If yes, in which context? Is this defined anywhere so we could add this information to the annotation?

For example, in Ile de France, the urban trains (RER) have a short code, that is shown to the public (on timetable, displays on the platform and in vehicles) because there are a lot of different branches and destination and passengers need to be sure what train they are jumping in. This code is short enough to be shown even on old displays.

OJP/OJP_JourneySupport.xsd
```diff
@@ -63,6 +63,11 @@
 		</xs:annotation>
 		<xs:sequence>
 			<xs:element ref="JourneyRef"/>
+			<xs:element name="PublicCode" type="xs:normalizedString" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Public code of the SERVICE JOURNEY.</xs:documentation>
```
sgrossberndt, line 68: As the element name "Public Code" has been chosen to be in line with NeTEx as @ue71603 wrote this should be reflected in the annotation like done in other elements, e.g.

```
<xs:element name="PublishedServiceName" type="InternationalTextStructure">
	<xs:annotation>
		<xs:documentation>Line name or service description as known to the public, f.e. "512", "S8" or "Circle Line" or "ICE 488".</xs:documentation>
	</xs:annotation>
</xs:element>
<xs:element name="TrainNumber" type="xs:normalizedString" minOccurs="0">
	<xs:annotation>
		<xs:documentation>Contains the TrainNumber description from NeTEx. If several TrainNumber types exist, use the commercial number.</xs:documentation>
	</xs:annotation>
</xs:element>
```

sgrossberndt (changes requested):

OJP/OJP_JourneySupport.xsd
ue71603: changed

Aurige (approved):

ue71603:
@sgrossberndt  Can you live with the clarification or still a problem?

sgrossberndt (approved):

PR #204, opened 2022-06-02
Include AllowedSystemId also in PlaceRef

ue71603:
....and perhaps we even should add it more generally depending on the algorithms.
The people working with EU spirit and HaCon/MENTZ should take a look.


JonaVBB:
Disclaimer: I may understand this imperfectly.
The role of the AllowedSystemId may differ. In EU-Spirit it is used in InitialInput to tell the RCC to which passive servers the request shall be passed. [In EU-Spirit the OJP services do not interact directly, but only via the RCC].
I understand that there may be an advantage to have the AllowedSystemId information available outside the InitialInput. Maybe you could elaborate a bit more what kind of procedure you have in mind. Do you want to have a chane to get an overview about available systems automatically before the actual request?

ue71603:
@JonaVBB  I think that in a distributed context it is sometimes necessary to know which system is the "owner/master" of the element (here place). In theory otherwise they might have the same id.  i would add a SystemId everywhere, where this distinction is necessary. may be also in TripInfoRequest. What do you think?

ue71603:
@JonaVBB  could you please comment?

herlitze (approved):

OJP/OJP_PlaceSupport.xsd
```diff
@@ -460,6 +460,11 @@
 					<xs:documentation>Public name of the place.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element name="AllowedSystemId" type="xs:normalizedString" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Used in distributed environments. e.g. EU-Spirit. If none is given, the place information request refers to all known systems (in EU-Spirit "passive servers"). If at least one is given, the place information request refers only to the given systems (in EU-Spirit "passive servers"). In EU-Spirit the system IDs were previously called "provider code". See https://eu-spirit.eu/</xs:documentation>
```
sgrossberndt, line 465: The text still says "If at least one is given" hinting there could be multiple but the unbounded has been removed recently. I don't know whether allowing only one or multiple `AllowedSystemId` elements is correct but the text should be accordingly

sgrossberndt (changes requested):

OJP/OJP_PlaceSupport.xsd
JonaVBB: From EU-Spirit's point of view "If at least one is given" is correct. Either the active system defines a list of allowed passive servers to be involved or leaves the list empty which means all systems may be involved.

ue71603: updated

skinkie (approved):

sgrossberndt (approved):

Issue #168, opened 2021-07-15
Inconsistency in PlaceStructure and PlaceTypeEnumeration


schlpbch:
![PlaceStructure-PlaceTypeEnumeration](https://user-images.githubusercontent.com/18540761/125785180-32ba3bd9-995a-4851-8007-0ddde605eada.GIF)

schlpbch:
Questions:

- Do "StopPlace" and "StopPoint" both map to "stop"  
- Does "GeoPosition" map to the  "coord" enum? Better: "geoPosition"

ue71603:
Addressed in: https://github.com/VDVde/OJP/pull/218

PR #218, opened 2022-07-20
Replace `coord` with `location` in `PlaceTypeEnumeration`

ue71603:
Enumeration updated . Fixes https://github.com/VDVde/OJP/issues/168


herlitze (dismissed):

OJP/OJP_Places.xsd
```diff
@@ -203,10 +203,23 @@
...
+			</xs:enumeration>
 			<xs:enumeration value="address"/>
 			<xs:enumeration value="poi"/>
-			<xs:enumeration value="coord"/>
+			<xs:enumeration value="coord">
+				<xs:annotation>
+					<xs:documentation>deprecated. Is replaced by geoPosition.</xs:documentation>
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="geoPosition">
```
skinkie, line 218: Wouldn't the transmodel "location" not be a better suited value here?
```suggestion
			<xs:enumeration value="location">
```

OJP/OJP_Places.xsd
```diff
@@ -203,10 +203,23 @@
...
+			</xs:enumeration>
 			<xs:enumeration value="address"/>
 			<xs:enumeration value="poi"/>
-			<xs:enumeration value="coord"/>
+			<xs:enumeration value="coord">
+				<xs:annotation>
+					<xs:documentation>deprecated. Is replaced by geoPosition.</xs:documentation>
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="location">
```
Aurige, line 218: Is an address a location ? I guess yes, but may then be explicit. If not "address" needs to be added

ue71603: @Aurige Where would we have to add it? In Transmodel? the problem is, that it all should e a PLACE not a LOCATION. An Address is a sort of PLACE: So I guess that what we are doing here, is ok. Or should we change something?

Aurige: My fault, nothing to add, _address_ is just above... I don't know how I missed it, sorry

OJP/OJP_Places.xsd
```diff
@@ -203,10 +203,23 @@
...
+				<xs:annotation>
+					<xs:documentation>stop covers StopPlace and StopPoint.</xs:documentation>
+				</xs:annotation>
+			</xs:enumeration>
 			<xs:enumeration value="address"/>
 			<xs:enumeration value="poi"/>
-			<xs:enumeration value="coord"/>
+			<xs:enumeration value="coord">
+				<xs:annotation>
+					<xs:documentation>deprecated. Is replaced by geoPosition.</xs:documentation>
```
sgrossberndt, line 215: `coord` is replaced by `geoPosition`, but there is no `geoPosition` in the `PlaceTypeEnumeration`.

What about `Deprecated. Use location instead.`

OJP/OJP_Places.xsd
```diff
@@ -203,10 +203,23 @@
...
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="location">
+				<xs:annotation>
+					<xs:documentation>New value for geoPosition.</xs:documentation>
```
sgrossberndt, line 220: What about `Geographic position consisting of coordinates. Replaces the deprecated value coord.`

sgrossberndt (changes requested):
Please try to add more meaningful commit messages so it is easier to identify which changes are actually done in the commit.

OJP/OJP_Places.xsd
ue71603: fixed

OJP/OJP_Places.xsd
ue71603: fixed

Aurige (approved):

sgrossberndt (approved):

trurlurl:
Documentation updated.

sgrossberndt (approved):

skinkie (approved):

herlitze (approved):

Issue #213, opened 2022-07-18
New OptimisationMethod shortestPath


ue71603:
In the area of IV and MIV routing this is also an optimisation.
Perhaps we should change OptimisaitonMethod to OptimisationMethodList and allow multiple results in one go.

E.g. IV shortestPath and fastest and leastCost.

PR #219, opened 2022-07-25
Optimisation method shortes path

ue71603:
fixes: https://github.com/VDVde/OJP/issues/213


trurlurl:
Documentation updated.

Issue #194, opened 2022-05-13
Discuss the influence AccessFacilities (from SIRI) in the OJP query and reply


skinkie:
The accessfacilities include information concerning station information for example. We must discuss the influence of SIRI in the OJP query and reply structures. For example step free when the escalator is broken.

ue71603:
Stefan: SituationExchange stuff. The why (because escalator is broken is the interesting part). We should be able to show in the Transferleg that the Accessibility is broken.  I always have first the static planning and then should be able to give the why.
Still send them the common reply, but tell, it is broken. Then we point to Siri.
Malte: Not only accessibility, but also realtime.
Stefan: This goes to the access facilities. 
Malte: If 4 trips are requested, what is responded.
Stefan:  I would get 8 responses
Malte:  With blocking E.g. the next may be 5 hours later.
Klaus: There is also a kind of "realtime information" related to accessability: there might be a e.g. elevator, but if it is actually broken/out of order, the trip might not possible or transfer times might be longer due to detours.
Malte: Implicit interval needed (first real-time result)
Matthias: User Preferences and the score may also play into it. 

Stefan: What would ideal trip reply do. 

Stefan will provide a some concept.  Work with André

ue71603:
André: I will create a PR with my proposal.

trurlurl:
This issue is followed up by PR#235 and can be closed. I don't seem to have the right permissions to do it.

PR #221, opened 2022-07-25
Update on User Preferences

ue71603:
USER PREFERENCE added with a lot of possible content. The attached paper is an introduction (not 100% aligned between PR and documentation yet. As I think this needs discussion, We will adapt pull request and specification at the same time.
Includes work done by  @markus-meier-sbb  and @herlitze
[2144934064_4af1fe6093104c4fbe434b74bba5cfb0-250722-1701-3540.pdf](https://github.com/VDVde/OJP/files/9182303/2144934064_4af1fe6093104c4fbe434b74bba5cfb0-250722-1701-3540.pdf)

### DIFF
--- a/OJP/OJP_Trips.xsd
+++ b/OJP/OJP_Trips.xsd
@@ -54,6 +54,7 @@
 			<xs:group ref="TripMobilityFilterGroup"/>
 			<xs:group ref="TripPolicyGroup"/>
 			<xs:group ref="TripContentFilterGroup"/>
+			<xs:element name="UserPreferences" type="UserPreferencesStructure" minOccurs="0" maxOccurs="1"/>
 			<xs:element name="FareParam" type="FareParamStructure" minOccurs="0"/>
 			<xs:element name="Extension" type="xs:anyType" minOccurs="0"/>
 		</xs:sequence>
@@ -147,6 +148,11 @@
 					<xs:documentation>Whether the result should include intermediate stops (between the passenger's board and alight stops).</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element name="IncludeSecondBestOptions" type="xs:boolean" default="false" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Whether second best options should be presented as well. Mainly important for dominated journeys or in the case of ContinuousLegs the second best route.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
 			<xs:element name="IncludeFare" type="xs:boolean" default="false" minOccurs="0">
 				<xs:annotation>
 					<xs:documentation>Whether the result should include fare information.</xs:documentation>
@@ -208,6 +214,17 @@
 			<xs:documentation>Trip response structure.</xs:documentation>
 		</xs:annotation>
 		<xs:sequence>
+			<xs:element name="Score">
+				<xs:annotation>
+					<xs:documentation>The score from User Preferences or the algorithm.</xs:documentation>
+				</xs:annotation>
+				<xs:simpleType>
+					<xs:restriction base="xs:float">
+						<xs:minInclusive value="0"/>
+						<xs:maxInclusive value="1"/>
+					</xs:restriction>
+				</xs:simpleType>
+			</xs:element>
 			<xs:element name="TripResponseContext" type="ResponseContextStructure" minOccurs="0">
 				<xs:annotation>
 					<xs:documentation>Context to hold trip response objects that occur frequently.</xs:documentation>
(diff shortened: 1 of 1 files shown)


ue71603:
Do we need some hasHeavyLuggage  boolean or something like that?

JonaVBB:
> Do we need some hasHeavyLuggage boolean or something like that?

Not sure. Would only make sense - from my point of view - if mist of ther requested systems would support such a request. Which then, refers to adequate time table data (vehicles and stations) to support specific searches.
Currently we only sipport wheelchairs, rollators etc. But maybe HeavyLuggage has the same implications than a rollator?

OJP/OJP_Trips.xsd
```diff
@@ -147,6 +148,11 @@
 					<xs:documentation>Whether the result should include intermediate stops (between the passenger's board and alight stops).</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element name="IncludeSecondBestOptions" type="xs:boolean" default="false" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Whether second best options should be presented as well. Mainly important for dominated journeys or in the case of ContinuousLegs the second best route.</xs:documentation>
```
skinkie, line 153: Second best is a vague term. If you want pluriformity in the search results, I would rather go for something specific like _IncludeAlternativeOptions_.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
 			<xs:element name="TripFare" type="TripFareResultStructure" minOccurs="0" maxOccurs="unbounded"/>
 		</xs:sequence>
 	</xs:complexType>
+	<xs:complexType name="UserPreferencesStructure">
```
skinkie, line 898: I miss a safety related parameter here. OpenTripPlanner for example has bicycle safety.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+							<xs:annotation>
+								<xs:documentation>fastest normal cycle route</xs:documentation>
+							</xs:annotation>
+						</xs:enumeration>
+						<xs:enumeration value="normalgreen">
+							<xs:annotation>
+								<xs:documentation>greenest normal cycle route</xs:documentation>
+							</xs:annotation>
+						</xs:enumeration>
+						<xs:enumeration value="normalcomfortrable">
```
skinkie, line 1038:
```suggestion
						<xs:enumeration value="normalcomfortable">
```

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+					<xs:restriction base="xs:string">
+						<xs:enumeration value="premium"/>
+						<xs:enumeration value="average"/>
+						<xs:enumeration value="basic"/>
+					</xs:restriction>
+				</xs:simpleType>
+			</xs:element>
+			<xs:element name="Sportiness" default="average" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>User's endurance and fitness for sportive, self-powered trip legs (walk, cycle, etd.).</xs:documentation>
```
skinkie, line 941:
```suggestion
					<xs:documentation>User's endurance and fitness for sportive, self-powered trip legs (walk, cycle, etc.).</xs:documentation>
```

skinkie:
As from the meeting I think I would like to see an addition in `OptimisationMethodEnumeration` stating userPreferences.

ue71603:
User Preference discussion 18.8.2022

•	Malte
•	Markus
•	André
•	David
•	Thomas
•	Stefan
•	Gisela
•	Matthias

Introduction by Markus
Update on User Preferences by ue71603 · Pull Request #221 · VDVde/OJP (github.com)
•	Malte: Metacriteria are ok. scoring of trip options is depending of the exceptation the hard part to discuss.
•	Malte: Rating on the client side may be too late.
•	Easy options in the back end
•	User preferences needed for scoring
•	
•	Stefan: How to deal with users
•	Malte: User profiles would be on the cllient side. We do the mapping on the client side. Multi-tab-approach used often (one for PT, one for sharing). 
•	Malte: We have abortion criteria for performance reasons
•	Stefan: How much do I have to relax the algorithm to meet the options. Some are defined as negative and some as positive. Different implementations give different results
•	Markus: that ist he case today, too. Wit
•	Stefan: Where would the preferences influence the score? Backend or client?
•	Markus: That is implementation stuff.
•	Stefan: OptimisationMethod userPreferences. Abortion criteria is more criteria and not time as usual
•	Malte: PT and IV are often separated. Currently only time is exchanged. For Stefan's approach need to be widened.
•	Malte: Reculaculation with relaxation is possible, but need to be strictly defined. It needs to be said, what is relaxed.
•	André fastest, most scienic, most accessible. Two searches done. 

Decisions: All user preferences need to be mappable to low – level functions, if needed add additonal low-level criteria. 
Decision: We add the User Preferences and it works
Decision: The explicitely set low level options  superseed the user preferences.

AccessFeature
Essential, Prefered  to be discussed.
Malte and Stefan did not like this.
Stefan, if this results in a score, what happens. Did I meet the criteria?

Decision: Multiple searches for accessibility. 
AccessFacilities
Discuss the influence AccessFacilities (from SIRI) in the OJP query and reply · Issue #194 · VDVde/OJP (github.com)

Decision: Combined request and the several requests with single optimisation.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+				<xs:annotation>
+					<xs:documentation>User's preference for scenic, touristic impressions on some parts of the trip.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="Peaceful" type="xs:boolean" default="false" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>User's preference for peacful transport (quiet travel compartements or low occupation).</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="TripReason" type="TripReasonEnumeration" minOccurs="0" maxOccurs="1">
```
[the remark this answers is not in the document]
Aurige: Ok, I saw it in the document ... sorry

OJP/OJP_Trips.xsd
ue71603: Malte: How far dominated?
Markus: Not only time. Price etc., but playing with strategy.
Stefan: So give me two is not good.
Malte: Even one that is worse in everything. Current algorithm throw out options. Truly dominated options, not throwing it out, would be a problem. 
Alexandre (Cityways): I worked on a PhD thesis. Distance defined, the results are interesting. Matchin user expectence. (chapter 5, https://tel.archives-ouvertes.fr/tel-01848737 ). A random display of trips that maximize diversity.
Malte: Not losing to much quality is important. 
Matthias: More diversity is the idea
Malte: Clearer annotation needed.
==> Annotation needs be better.
==> IncludeAlternativeOptions will be renamed.

OJP/OJP_Trips.xsd
ue71603: Stefan: safe vs fast. safety was more of not getting robbed below. 
Markus: Extrasafe is not really only about that. 
Stefan: What do you meand for safe? In OTP it is traffic safety. 
Markus: I meant accidents and crime.
Stefan: I suggest to split it into 
Safety:
* traffic safety. TrafficSafe
* feeling safe   xxx

ue71603: Alexandre: https://crimeandplace.com/

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+						</xs:enumeration>
+						<xs:enumeration value="difficult">
+							<xs:annotation>
+								<xs:documentation>E.g. white-blue-white hiking signs</xs:documentation>
+							</xs:annotation>
+						</xs:enumeration>
+					</xs:restriction>
+				</xs:simpleType>
+			</xs:element>			
+			<xs:element name="CyclingProfile" default="normalfast" minOccurs="0" maxOccurs="1">
```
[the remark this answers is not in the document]
ue71603: Will adapt

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+				<xs:annotation>
+					<xs:documentation>User's preference for an extra high level of safety.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="ExtraReliable" type="xs:boolean" default="false" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>User's preference for an extra high reliability (low probabilty of delays, cancellations, etc.).</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="ExtraAccessible" type="xs:boolean" default="false" minOccurs="0" maxOccurs="1">
```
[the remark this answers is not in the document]
ue71603: Stefan: I don't think we need to have it more than the other accessibility description. 
Markus: More emphasis.
Malte: All bases are already covered (like NoStairs).
Stefan: Probably another OptimisationMethod "Accessible"
Markus:  Not changing too much, extra time for changes
Malte: One request with multiple optimisation methods. This makes things more vague and will add more problems on this. We should try to minimize there.
Markus: I think next step would be to write the chapter. It should be more hints as the trip planner knowing more.

ue71603:
Malte: How do we get to a similar understanding to these user preferences? Same interpretations from distributed systems.?
Markus: Simply adding this to trip request. All should get the user preferences and try to do the best job. It works that way today. Also currently they work differently. 
Malte: It should be a meta parameter set. Would it still send the meta set.
Markus: It should do both. 
Matthias: We could and perhaps should a chapter on user preferences and interpretations to the standard.
André: There will also be a translation from high-level and low-level preferences. Will we need to implement the higher level, when it can be expressed the other way. 
Malte: More vague terms should be avoided. The goal was to have simple way for a client to transmit some user settings.
==> chapter to be written once it is clear, what is defined.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
 		</xs:sequence>
 	</xs:complexType>
+	<xs:complexType name="UserPreferencesStructure">
+		<xs:annotation>
+			<xs:documentation>A collection of user preferences to allow the algorithm to narrow-down multimodal trip combinations to those that may be suitable for the user.</xs:documentation>
+		</xs:annotation>
+		<xs:sequence>
+			<xs:element name="LowPrice" default="average" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>User's preference that the trip has a low price.</xs:documentation>
```
markus-meier-sbb, line 905: [This extra text could be added to give more hints about the meaning and how it should be implemented]
If important, the trip planner should prioritize options (modes, lines, operators) known for low price, such as bus or coach.
If unimportant, also consider options which are usually more epensive, such as taxi or ridehailing, and also for longer distances.
E.g., take a taxi to the central station instead of a PT ride with transfers.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+						<xs:enumeration value="important"/>
+						<xs:enumeration value="average"/>
+						<xs:enumeration value="unimportant"/>
+					</xs:restriction>
+				</xs:simpleType>
+			</xs:element>
+			<xs:element name="Speed" default="average" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>User's preference about the speed/short duration of the trip.</xs:documentation>
+				</xs:annotation>
```
markus-meier-sbb, line 918: _[This extra text could be added to give more hints about the meaning and how it should be implemented]_
If important, duration should be the top criterion and only the fastest connections should be considered.
If unimportant, duration should have little or no influence on the trip planning.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+				<xs:simpleType>
+					<xs:restriction base="xs:string">
+						<xs:enumeration value="important"/>
+						<xs:enumeration value="average"/>
+						<xs:enumeration value="unimportant"/>
+					</xs:restriction>
+				</xs:simpleType>
+			</xs:element>
+			<xs:element name="Comfort" default="average" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
```
markus-meier-sbb, line 928: _[This extra text could be added to give more hints about the meaning and how it should be implemented]_
If premium, prioritize few and short changes, and modes known for convenience, especially taxi.
Avoid self-powered modes and urban transit.
If basic, allow for more transfers and allow all modes.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+						<xs:enumeration value="premium"/>
+						<xs:enumeration value="average"/>
+						<xs:enumeration value="basic"/>
+					</xs:restriction>
+				</xs:simpleType>
+			</xs:element>
+			<xs:element name="Sportiness" default="average" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>User's endurance and fitness for sportive, self-powered trip legs (walk, cycle, etc.).</xs:documentation>
+				</xs:annotation>
```
markus-meier-sbb, line 942: _[This extra text could be added to give more hints about the meaning and how it should be implemented]_
If high, allow for 200 % longer legs with self-powered modes and assume 50 % faster speed on those modes.
If low, restrict to 50 % shorter walking legs, assume 50 % lower walking speed and exclude
other self-powered/micromobility modes.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+				<xs:simpleType>
+					<xs:restriction base="xs:string">
+						<xs:enumeration value="high"/>
+						<xs:enumeration value="average"/>
+						<xs:enumeration value="low"/>
+					</xs:restriction>
+				</xs:simpleType>
+			</xs:element>
+			<xs:element name="EnvironmentalSafety" default="average" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
```
markus-meier-sbb, line 952: _[This extra text could be added to give more hints about the meaning and how it should be implemented]_
If high, favour "green" modes/lines such as bike sharing and (electric) trains, avoid or restrict modes/lines
known for higher CO2 emissions such as (conventional) taxi, ridehailing or coach.
If low, basically ignore (same as average).
(Probably better to turn it into a true/false option)

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+				<xs:simpleType>
+					<xs:restriction base="xs:string">
+						<xs:enumeration value="high"/>
+						<xs:enumeration value="average"/>
+						<xs:enumeration value="low"/>
+					</xs:restriction>
+				</xs:simpleType>
+			</xs:element>
+			<xs:element name="WeatherProtection" type="xs:boolean" default="false" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
```
markus-meier-sbb, line 964: _[This extra text could be added to give more hints about the meaning and how it should be implemented]_
If true, micromobility should be excluded and walks for the first/last leg avoided,
e.g. use a bus or taxi even for a short distance.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+					<xs:documentation>User's preference not be exposed to eventual weather hazards, thus, to avoid walks and micromobility.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="ExtraSafe" type="xs:boolean" default="false" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
```
markus-meier-sbb, line 969: _[This extra text could be added to give more hints about the meaning and how it should be implemented]_
If true, certain modes, lines or zones/districts known for lower safety, i.e. higher risk of accidents and crime,
may be avoided, others may be preferred. This may depend on the actual, local or time of day situation.
E.g. bike or scooter may be considered unsafe in some cities/districts while safe in others.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+					<xs:documentation>User's preference for an extra high level of safety.</xs:documentation>
+				</xs:annotation>
```
markus-meier-sbb, line 971: _[This extra text could be added to give more hints about the meaning and how it should be implemented]_
If true, certain modes, lines or zones/districts known for lower safety, i.e. higher risk of accidents and crime,
may be avoided, others may be preferred. This may depend on the actual, local or time of day situation.
E.g. bike or scooter may be considered unsafe in some cities/districts while safe in others.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+			</xs:element>
+			<xs:element name="ExtraReliable" type="xs:boolean" default="false" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>User's preference for an extra high reliability (low probabilty of delays, cancellations, etc.).</xs:documentation>
+				</xs:annotation>
```
markus-meier-sbb, line 976: _[This extra text could be added to give more hints about the meaning and how it should be implemented]_
If true, modes known for their (un)reliability may be avoided/preferred, and extra time added for transfers.
This may depend on the actual, local or time of day situation,
based on punctuality statics, traffic jam statistics or rush hours.
E.g. taxis in a given city might be known to be unreliable during at 8-10 and 16-19 hours, otherwise reliable.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+			</xs:element>
+			<xs:element name="ExtraAccessible" type="xs:boolean" default="false" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>User's preference for perfect accessibility (lifts, ramps, no stairs, etc.), including extra transfer time, short transfer distances, etc.. May be superseeded by EPIAP updates. TODO</xs:documentation>
+				</xs:annotation>
```
markus-meier-sbb, line 981: _[This extra text could be added to give more hints about the meaning and how it should be implemented]_
If true, few and short transfers are preferred, stations/hubs with insufficient accessibility avoided.
(Based on the assumption that these same measures help in most cases,
regardless of the details of the actual impairment.
E.g. a station well equiped for wheelchairs will typically also be well equiped for visually impaired.)

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+			</xs:element>
+			<xs:element name="Scenic" type="xs:boolean" default="false" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>User's preference for scenic, touristic impressions on some parts of the trip.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
```
markus-meier-sbb, line 987: _[This extra text could be added to give more hints about the meaning and how it should be implemented]_
If true, certain lines or routes known for their scenicness may be prioritized.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+			<xs:element name="Peaceful" type="xs:boolean" default="false" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>User's preference for peacful transport (quiet travel compartements or low occupation).</xs:documentation>
+				</xs:annotation>
```
markus-meier-sbb, line 991: _[This extra text could be added to give more hints about the meaning and how it should be implemented]_
If true, modes/lines offering extra privacy should be preferred.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+			</xs:element>
+			<xs:element name="TripReason" type="TripReasonEnumeration" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>A classification of a passenger's reason for undertaking a TRIP PATTERN (TRIP REASON from Transmodel).</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="WalkingProfile" default="normal" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>Users walking profile (especially for hiking)</xs:documentation>
+				</xs:annotation>
```
markus-meier-sbb, line 1001: May be somewhat exotic or rarely used preference. Is it useful in the context of the other user preferences? Are these colour codes standardized in Europe? Most countries will have none of these.

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+			<xs:element name="Peaceful" type="xs:boolean" default="false" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>User's preference for peacful transport (quiet travel compartements or low occupation).</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="TripReason" type="TripReasonEnumeration" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>A classification of a passenger's reason for undertaking a TRIP PATTERN (TRIP REASON from Transmodel).</xs:documentation>
+				</xs:annotation>
+			</xs:element>
```
markus-meier-sbb, line 997: This classification may have overlaps with the ones above. Needs to be discussed whether this is useful in combination with the above preferences, or as an alternative, or even a complete, sufficient replacement of the above criterions?

OJP/OJP_Trips.xsd
```diff
@@ -878,6 +895,172 @@
...
+						<xs:enumeration value="low"/>
+					</xs:restriction>
+				</xs:simpleType>
+			</xs:element>
+			<xs:element name="EnvironmentalSafety" default="average" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>User's preference about climate/environmental-friendliness of the trip (option 'low' might have no effect).</xs:documentation>
+				</xs:annotation>
+				<xs:simpleType>
+					<xs:restriction base="xs:string">
```
Aurige, line 956: May be define common types for enumeration to avoid multiple definition of the same values (here high/average/low, but this happens several times)

ue71603:
is now implemented in https://github.com/VDVde/OJP/pull/271

trurlurl:
Removed *documentation* label, documentation will be done for the follow-up PR.

PR #223, opened 2022-08-08
Add trip optimisation method `leastDistance`

skinkie:
Better than #219


ue71603 (approved):

ue71603:
replaced with: https://github.com/VDVde/OJP/pull/242

Issue #222, opened 2022-08-05
EntitlementProduct should be structured differently


ue71603:
EntilementProduct probably is underdefined.
Either ist should be EntitlementProductRef or better it should structured like ZOnesAlreadyPaid

Containing a 
* FareAuthorityRef and a
* EntitlementProductRef

What are the other's opinions?
![image](https://user-images.githubusercontent.com/24227470/183040356-5ca354bb-f232-4241-8e28-b43a1cbc106c.png)

ue71603:
Stefan: From catalog. Not user specific. 
Dirk: I have a student card from PTA.   DB:StudentCard

EntitlementProduct
- FareAuthorityRef  = OeVCH
- EntitlementProductRef  = 818231
- EntitlementProductName = Halfprice

Usually selected from a catalog.
Will ask Andreas S., if such a catalogue exist.

Decision: Restructure and document (and we have to see where the catalogue comes from)

ue71603:
[09:26] Dohmen, Claus | VDV
There may be additional attributes of a entitlement product necessary to use it in a fare calculation, e.g. validity period

[09:28] Dohmen, Claus | VDV
This could not be covered by a catalogue, because it is am attribute of the individual card rather than of the card type (like e.g. discount level)

PR #229, opened 2022-08-11
Extend definition of entitlement products

ue71603:
The `FarePassengerStructure` could contain multiple `EntitlementProductRef`s before, which did not contain much information, just the ID of the entitlement product. With this change the `FarePassengerStructure` may contain a `EntitlementProductList` instead consisting of  `EntitlementProduct`s, which allows to define for each entitlement product:
* FareAuthorityRef
* EntitlementProductRef
* ValidityPeriod
* EntitlementProductName

This improves the `EntitlementProduct` definition to match elements like `ZonesAlreadyPaid`.

Fixes https://github.com/VDVde/OJP/issues/222


OJP/OJP_FareSupport.xsd
```diff
@@ -408,9 +432,9 @@
 					</xs:annotation>
 				</xs:element>
 			</xs:choice>
-			<xs:element name="EntitlementProduct" type="EntitlementProductRefStructure" minOccurs="0" maxOccurs="unbounded">
+			<xs:element name="EntitlementProducts" type="EntitlementProductListStructure" minOccurs="0" maxOccurs="unbounded">
```
sgrossberndt, line 435: The structure allowed multiple `EntitlementProduct` elements to exist beneath each other before the change due to the `maxOccurs="unbounded"`. So you are now introduction an explicit list element to group them.
Was the motivation only to have multiple EntitlementProducts and you overlooked the `maxOccurs="unbounded"` or was it intentional to have them in a list? If for the first option this change would not be necessary, if for the second option the `maxOccurs="unbounded"` on the list must be removed because otherwise there could be multiple groups.

OJP/OJP_FareSupport.xsd
```diff
@@ -217,9 +230,20 @@
 		<xs:annotation>
 			<xs:documentation>Reference to a a precondition to access a service or to purchase a FARE PRODUCT issued by an organisation that may not be a PT operator (eg: military card, concessionary card, etc).</xs:documentation>
 		</xs:annotation>
-		<xs:simpleContent>
-			<xs:extension base="EntitlementProductCodeType"/>
-		</xs:simpleContent>
+		<xs:sequence>
```
sgrossberndt, line 233: I am wondering whether with this change the name `EntitlementProductRefStructure` is still valid or if it should be called `EntitlementProductStructure` now (without the `Ref`) as it contains several elements itself and no longer is a sole reference.

sgrossberndt (changes requested):

OJP/OJP_FareSupport.xsd
ue71603: Changed that as well

OJP/OJP_FareSupport.xsd
sgrossberndt, line 435: You are now introducing an explicit list element `EntitlementProducts` which will contain the `EntitlementProduct`s. Why does this element `EntitlementProducts` still have `maxOccurs="unbounded"`?
This would allow multiple `EntitlementProducts` lists beneath each other, each containing `EntitlementProduct` items.

These are the options to solve this:
1. Remove `maxOccurs="unbounded"` from `EntitlementProducts`. Now only one list may exist.
2. It is intentional that multiple `EntitlementProducts` elements each with `EntitlementProduct` items may exist, but then an explanation is missing why.
3. We do not need an explicit `EntitlementProducts` element to group the `EntitlementProduct` items at all. `EntitlementProduct` is not renamed to `EntitlementProducts` and kept with `maxOccurs="unbounded"`, only the type changes from `EntitlementProductRefStructure` to `EntitlementProductStructure`. `EntitlementProductListStructure` is removed and its annotation moved back to `EntitlementProduct`.

sgrossberndt (changes requested):
What is meant by `Improving Entitlement Product to match others`. Which "others"?

OJP/OJP_FareSupport.xsd
ue71603: Did 1. Hope this is ok for everybody

ue71603:
> What is meant by `Improving Entitlement Product to match others`. Which "others"?

updated title and description of pr

sgrossberndt (dismissed):

sgrossberndt:
https://htmlpreview.github.io/?https://github.com/VDVde/OJP/blob/Improving_Entitlement_Product/docs/generated/OJP.html#EntitlementProductCodeType

sgrossberndt (approved):

herlitze (approved):

skinkie (approved):

PR #215, opened 2022-07-20
Simplify element names of IDs

ue71603:
Fixes https://github.com/VDVde/OJP/issues/181 https://github.com/VDVde/OJP/issues/179

PR #230, opened 2022-08-11
Rename TripLeg to Leg

ue71603:


ue71603:
Fixes @180

ue71603:
will do all examples, when content is ok for group.

sgrossberndt:
OJP_FareSupport.xsd has a `TripLegRangeGroup` element referring to the elements `FromTripLegIdRef` and `ToTripLegIdRef` from OJP_Fare,xsd. Should those be renamed too?

ue71603:
Discussion
•	#215
•	#230
 
They both would align things more to what it should be. However, (a) we don’t do it everywhere and we should move the id to an attribute in that case (b) in breaks a lot of implementations (c) It does not really simplify anything in a heavy way. I agree: It is hygiene, but I am not sure, if we should do it.
Please advise. For now I will leave them alone

skinkie:
This is a breaking change... is this for 1.1.

ue71603:
> This is a breaking change... is this for 1.1.

It is for 2.0

skinkie (approved):

herlitze (approved):

AndreasAtSBB (approved):
Checkt also the diff inside the HTML files.

trurlurl (approved):

skinkie (dismissed):

sgrossberndt (changes requested):
OJP_FareSupport.xsd has a TripLegRangeGroup element referring to the elements FromTripLegIdRef and ToTripLegIdRef from OJP_Fare,xsd. Shouldn't those be renamed too?

skinkie (approved):

sgrossberndt (approved):

skinkie (approved):

PR #233, opened 2022-08-23
Improve title attributes in documentation

### LINKED ISSUE #159: Negative Multiplicity in HTML
Strangely, sometimes the multiplicity are negative in the HTML generated:

NegativeMulticipliesInHTML

### LINKED ISSUE #172: Resolve the infinite loop in the documentation
As brought up in #165 by @schlpbch. Not a show stopper, but we may think about a solution about it.

> image
> This suggests that PointOfInterestRef is of type PointOfInterestRef which would be an infinite loop.

@jmaerki replied:
> Yes, that's a consequence of the documentation rules where I remove the "Structure" Suffix. How would you expect this to be resolved? I can't just remove the "Ref" at the end because that would create another conflict between "PointOfInterestRef", "PointOfInterestRefStructure" and "PointOfInterest". Same for "TopographicPlace*". The only way out is to remove the rule that removes the "Structure" Suffix.

sgrossberndt:
* "title" attribute with the full structure element name on references (Fixes #172)
* "title" attribute with explanation of cardinality. (Fixes #159)


ue71603 (dismissed):

sgrossberndt:
Rebased only

ue71603 (dismissed):

sgrossberndt:
title attributes added: https://htmlpreview.github.io/?https://github.com/VDVde/OJP/blob/changes_for_v1.1/docs/generated/OJP.html
before: https://htmlpreview.github.io/?https://github.com/VDVde/OJP/blob/088b65f13e49f20edf8a89448ea35bb2e4227d99/docs/generated/OJP.html

herlitze (approved):

ue71603 (approved):

Issue #145, opened 2021-01-06
Extending PathGuidanceSection for accessibility/levels


ue71603:
We have studied indoor routing through Transferleg. 
There it should be possible to tell people to move with an elevator to level -1 (and the like).
I didn't see that yet. 

Example:
![image](https://user-images.githubusercontent.com/24227470/103746136-e230c080-5000-11eb-96c3-35d2df47ddcb.png)

Should I create a description and a pull request? Or does there already exist work?

StefanEngelhardt:
Our passive server is delivering those informations in a localized way in PathGuidanceSection/TrackSection/RoadName.

Examples:
                      <ojp:RoadName>Treppe abwärts zur Ebene -1</ojp:RoadName>
                      <ojp:RoadName>Rolltreppe abwärts zur Ebene -3</ojp:RoadName>

ue71603:
concrete proposal

ue71603:
FromLevelRef ToLevelRef in PathLink would be a possible place

ue71603:
Also add some EquipmentStatus from NeTex to PathLink

ue71603:
<xsd:simpleType name="EquipmentStatusEnumeration">
		<xsd:annotation>
			<xsd:documentation>Allowed values for status of EQUIPMENT.</xsd:documentation>
		</xsd:annotation>
		<xsd:restriction base="xsd:NMTOKEN">
			<xsd:enumeration value="unknown"/>
			<xsd:enumeration value="available"/>
			<xsd:enumeration value="notAvailable"/>
		</xsd:restriction>
	</xsd:simpleType>

trurlurl:
This issue is followed up by PR#234 and can be closed. I don't seem to have the right permissions to do it.

trurlurl:
Follow-up in PR #239.

PR #234, opened 2022-08-23
Allows for level information on PathLinks

trurlurl:
Allows for level information on PathLinks. This PR responds to #145.

#145 asks for more data to tell a passenger, for instance, to use an elevator leading from level "A" to level "U1". This is achieved by adding FromLevel, ToLevel, LevelNameStructure.

Typical attributes useful for guidance in that kind of use cases (example):
PathLink.Transition: down 
PathLink.FromLevel.PublicCode: A  (new)
PathLink.ToLevel.PublicCode: U1  (new)

Two other PRs related to this will follow soon - dealing with real-time information about access equipment (#194) and additional accessiblity information.

### DIFF
--- a/OJP_Trips.xsd
+++ b/OJP_Trips.xsd
@@ -795,29 +795,29 @@
 					<xs:documentation>Number how often the access feature occurs in this PathLink</xs:documentation>
 				</xs:annotation>
 			</xs:element>
-			<xs:element name="FromLevel" type="LevelNameStructure" minOccurs="0">
+			<xs:element name="From" type="PathLinkEndStructure" minOccurs="0">
 				<xs:annotation>
 					<xs:documentation>Designations of the level where the PathLink starts.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
-			<xs:element name="ToLevel" type="LevelNameStructure" minOccurs="0">
+			<xs:element name="To" type="PathLinkEndStructure" minOccurs="0">
 				<xs:annotation>
 					<xs:documentation>Designations of the level where the PathLink ends.</xs:documentation>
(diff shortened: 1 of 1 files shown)


Aurige:
Shouldn't PathLinks have a "from" and a "to" where the level information could be stored (as in NeTEx) ?

trurlurl:
@Aurige Yes, perhaps we should have From and To where additional information can be added. What's the information one might want? Coordinates of the centroid, whether the coordinates are the centroid of a larger space or a narrower goal like an equipment place (to use in maps), a polygon of the space, the name if it is sufficiently useful (how to determine?), an id ?

If we want anything like that, then I should rename the proposed elements (I'm going to do that in any case) such that one can use
PathLink.From.LevelName (instead of PathLink.FromLevel.Name)
PathLink.To.LevelPublicCode
PathLink.To.Coordinates

I think it depends on how PathGuidance is intended to be put into work. Geographical information doesn't seem to play an important role. But how can the link be done to map planners? Does the complete picture exist somewhere, or do we have to think this through?

Aurige:
Ok for me. I would just add an **Id** to the **PathLinkStructure** so you can relate it to some map data, as you resquest, or to some perviously exchanged NeTEx dataSet.

ue71603 (commented):
OK, but should be based on changes_for_v1.1 and be merged into that.

trurlurl:
If this needs to be merged into changes_for_v1.1 instead of master - I don't seem to have the necessary permissions.
Neither can I assign the item to reviewers.

Aurige (approved):

ue71603:
replaced by #239

PR #235, opened 2022-08-23
Add AccessFeatureStatus on PathLink to indicate a broken access equipment.

trurlurl:
AccessFeatureStatus on a PathLink to indicate a broken access equipment.

#194 asks for data to tell whether, e.g., an elevator is broken, but to nevertheless show the route so that the passenger knows that this route would exist but is currently not feasible.

EquipmentStatusEnumeration has been added from NeTEx to indicate the status of an access equipment. The following would indicate that the elevator is broken:

PathLink.AccessFeatureStatus: notAvailable  (new)

The broken elevator makes the complete TransferLeg unfeasible - this should be flagged in the TransferLeg (e.g., TransferLeg.Feasibility : accessEquipmentOutOfService) or/and at the level of the Trip. 
What solution would you prefer? 

To also handle the case that the TranferLeg is not feasible due to a SituationFull (can that happen?), this could be indicated by:
TransferLeg.Feasibility: seeSituations

### DIFF
--- a/OJP_Trips.xsd
+++ b/OJP_Trips.xsd
@@ -775,14 +775,24 @@
 			<xs:enumeration value="downAndUp"/>
 		</xs:restriction>
 	</xs:simpleType>
+	<xs:simpleType name="EquipmentStatusEnumeration">
+		<xs:annotation>
+			<xs:documentation>Allowed values for status of EQUIPMENT.</xs:documentation>
+		</xs:annotation>
+		<xs:restriction base="xs:string">
+			<xs:enumeration value="unknown"/>
+			<xs:enumeration value="available"/>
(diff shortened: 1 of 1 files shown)


ue71603 (changes requested):
Shouldn't this be merged agains changes_for_1.1? Otherwise it is fine.

trurlurl:
@skinkie Could you please review this? Also:
If this needs to be merged into changes_for_v1.1 rather than master - I don't seem to have the necessary permissions to change that.
Neither can I assign the item to reviewers. Do I just not find how to do it, or is it OK like that?

ue71603:
Fixed in #238

PR #236, opened 2022-08-23
Update OJP_Trips.xsd

trurlurl:
Additional AccessFeatureTypes and new AccessibilityFeatures attribute.

In the context of PR#234 and #235, I thought it would be good to add accessibility-related data like the following:

PathLink.AccessibilityFeatures: visualSigns, elevatorAccousticAnnouncements, elevatorTactileButtons, etc. This allows for returning accessiblity information attached to PathLinks.

I made AccessFeatureTypes more consistent and complete regarding the possible access equipments by adding singleStep, travelator, etc.

I also changed lift to elevator, to be consistent with NeTEx.


ue71603:
replaced with #237

PR #239, opened 2022-08-26
Allow level information on PathLinks

ue71603:
Allows for level information on PathLinks. This PR responds to https://github.com/VDVde/OJP/issues/145.

https://github.com/VDVde/OJP/issues/145 asks for more data to tell a passenger, for instance, to use an elevator leading from level "A" to level "U1". This is achieved by adding FromLevel, ToLevel, LevelNameStructure.

Typical attributes useful for guidance in that kind of use cases (example):
PathLink.Transition: down
PathLink.FromLevel.PublicCode: A (new)
PathLink.ToLevel.PublicCode: U1 (new)

Two other PRs related to this will follow soon - dealing with real-time information about access equipment (https://github.com/VDVde/OJP/issues/194) and additional accessiblity information.

Replaces #234


Aurige (approved):

OJP/OJP_Trips.xsd
```diff
@@ -743,6 +743,38 @@
 					<xs:documentation>Number how often the access feature occurs in this PathLink</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element name="From" type="PathLinkEndStructure" minOccurs="0">
```
sgrossberndt, line 746: This structure makes it possible to only supply a `From`, but no `To`. Maybe add a group consisting of exactly one `From` and one `To` element and give the group `minOccurs="0"` instead?

OJP/OJP_Trips.xsd
```diff
@@ -743,6 +743,38 @@
...
+				<xs:annotation>
+					<xs:documentation>Public identifier of the level as found on elevators and signs.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="LevelName" type="InternationalTextStructure" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Official name of the level.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="id" type="xs:normalizedString" minOccurs="0">
```
sgrossberndt, line 773: This lowercase `"id"` is unusual, I think our element names always start with an uppercase letter.

sgrossberndt (changes requested):
Instead of repeating `Level` in `From->LevelPublicCode` and `From->LevelName` and having `From->Id` what about naming `From` `FromLevel` and `To` `ToLevel` and thus having `FromLevel->PublicCode`, `FromLevel->Name`, `FromLevel->Id` instead? This `PublicCode` part would be more consistent with the just merged [simplified element names](https://github.com/VDVde/OJP/commit/a64fb8a694164ea072cc7c58480e50b531328d0f)

OJP/OJP_Trips.xsd
trurlurl: @sgrossberndt Of course, you're right. I'll change it ...

OJP/OJP_Trips.xsd
trurlurl: @sgrossberndt  The structure also makes it possible to supply only a *To*. But wouldn't that be welcome? - When giving directions, it is usually sufficient to provide information only for the *To* (for instance the PublicCode of the target level when using an elevator). 
Regarding your previous comment: The idea of the Id was not to point to a *Level*, but more generally to an element at the end of a PathLink - for instance a NeTEx AccessSpace or an EquipmentPlace (this would later allow for retrieving additional information, e.g., the coordinates of the EquipmentPlace of the elevator). Thus, *ToLevel*->*Id* would not be what was intended. That said, I think it is better to stay with the proposed solution. What do you think?

OJP/OJP_Trips.xsd
ue71603: changed it to Id.

trurlurl (approved):
OK. I think the change requested by @sgrossberndt is obsolete, see the first "show comment" in the conversation tab.

OJP/OJP_Trips.xsd
```diff
@@ -743,6 +743,38 @@
 					<xs:documentation>Number how often the access feature occurs in this PathLink</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element name="From" type="PathLinkEndStructure" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Designations of the level where the PathLink starts.</xs:documentation>
```
sgrossberndt, line 748: Designations of level and place where this PathLink starts?

OJP/OJP_Trips.xsd
```diff
@@ -743,6 +743,38 @@
...
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="To" type="PathLinkEndStructure" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Designations of the level where the PathLink ends.</xs:documentation>
```
sgrossberndt, line 753: Designations of level and place where this PathLink ends?

OJP/OJP_Trips.xsd
```diff
@@ -743,6 +743,38 @@
...
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="LevelName" type="InternationalTextStructure" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Official name of the level.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="Id" type="xs:normalizedString" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Id of the element at the From or To end of the PathLink (typically a PLACE).</xs:documentation>
```
sgrossberndt, line 775: Id of the element at this end of the PathLink (typically a PLACE, e.g. elevator)

sgrossberndt (changes requested):

OJP/OJP_Trips.xsd
trurlurl: @sgrossberndt Your're right - this is a better description of the structure's purpose since the addition of the Id attribute.

OJP/OJP_Trips.xsd
trurlurl: @sgrossberndt Your're right - this is a better description of the structure's purpose since the addition of the Id attribute.

OJP/OJP_Trips.xsd
trurlurl: @sgrossberndt OK - but perhaps without 'e.g. elevator'? I think the Id would typically point at the EquipmentPlace of the elevator (which has coordinates), not directly at the LiftEquipment. -> or: 'e.g., where the elevator is located' ?

OJP/OJP_Trips.xsd
ue71603: changed

OJP/OJP_Trips.xsd
ue71603: changed

OJP/OJP_Trips.xsd
```diff
@@ -743,6 +743,38 @@
...
+			<xs:element name="To" type="PathLinkEndStructure" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Designations of the level where the PathLink ends.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+		</xs:sequence>
+	</xs:complexType>
+	<xs:complexType name="PathLinkEndStructure">
+		<xs:annotation>
+			<xs:documentation>Designations of a floor.</xs:documentation>
```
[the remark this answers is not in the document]
ue71603: added level

OJP/OJP_Trips.xsd
ue71603: integrated your discussion

trurlurl:
Documentation updated accordingly.

trurlurl (dismissed):

Aurige (approved):

skinkie (dismissed):

sgrossberndt (dismissed):

ue71603:
We will use kilograms. Also use float

sgrossberndt (approved):

PR #242, opened 2022-09-12
add leastDistance as OptimisationMethodEnumeration

ue71603:
replaces: https://github.com/VDVde/OJP/pull/223


herlitze (approved):

sgrossberndt (approved):

sgrossberndt:
https://htmlpreview.github.io/?https://github.com/VDVde/OJP/blob/OptimisationMethod_extended/docs/generated/OJP.html

The annotation "Least distance in meter. Mostly used for ALTERNATIVE MODE OF OPERATION" will not be displayed in the HTML

ue71603:
@sgrossberndt  What should I do to correct this?

sgrossberndt:
I approved the change. I don't see an easy option to fix this. The current HTML definition simply does not display annotations for constants. The annotation is for now just visible in the XSD source. I just wanted to give this hint so other reviewers know why it is not displayed.

ue71603:
Thanks for the 

> I approved the change. I don't see an easy option to fix this. The current HTML definition simply does not display annotations for constants. The annotation is for now just visible in the XSD source. I just wanted to give this hint so other reviewers know why it is not displayed.

@skinkie  It might be rather difficult to extend this part in the XSL. What do you think, we leave it at that? So the detailed description of the ENUM is only in the XSD? The values are in the doc.

Issue #15, opened 2018-09-03
Add refinement service to OJP from TRIAS


sgrossberndt:
[Low priority](https://github.com/VDVde/TRIAS/tree/Version1.4)
From Trias the TripRefineRequest should be copied. 

The other refinements I would skip for the time being. 

The goal of the refinement request could be:
* more information about certain legs (e.g. real-time, fares, more details, situations). 

In a distributed environment a trip planner may calculate based on timetables and then use refinement to update the trips from different services.

Some documentation could be copied from Trias for the doc.

sgrossberndt:
This service from [TRIAS v1.2](https://github.com/VDVde/TRIAS) would be a useful addition to be ported to OJP 2.0.

BertRadi:
Will be postponed to OJP 2.0

ue71603:
@sgrossberndt @skinkie 
Should we include a TripRefinementRequest? For the other's I don't see really a use case.
In TripRefinement e.g. a responding system may add real time information. What do you think?

skinkie:
https://www.vdv.de/431-2sdes-v-1-3.pdfx (page 149) defines what refinement means.

ue71603:
* add realtime
* add fares
* has only references and must refine with product information
* enrichment
concept then PR
More requests needed.

ue71603:
mergt

PR #247, opened 2022-09-22
TripRefineRequest

trurlurl:
PR for issue #15.

### DIFF
--- a/OJP/OJP_Requests.xsd
+++ b/OJP/OJP_Requests.xsd
@@ -194,4 +194,31 @@
 			</xs:extension>
 		</xs:complexContent>
 	</xs:complexType>
+	<xs:annotation>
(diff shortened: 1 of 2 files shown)


ue71603 (approved):

ue71603:
Do we need a flag that says, that new legs may be created? E.g. an on-demand leg is given between origin-destination. The other legs are not only to be recalculated, but can be newly made. E.g a multleg public transport part. Also: Do we allow in the refinements that some stops can be recalculated even, when they are not explicitly mentioned? E.g. I have calculated a trip. I want to have on the first mile a scooter leg. Probably I should add a  walking leg at the beginning (which is new) and also I may want to reschedule the public transport part. What does the group think? We can do this as a second PR.

skinkie (approved):

PR #267, opened 2022-11-26
Rename FareProductCode to FareProductId

ue71603:
Should be aligned with type


OJP/OJP_FareSupport.xsd
```diff
@@ -255,9 +255,9 @@
 			<xs:documentation>[related to FARE PRODUCT in TM and NeTEx] different FARE PRODUCTs that may be available with related information</xs:documentation>
 		</xs:annotation>
 		<xs:sequence>
-			<xs:element name="FareProductId" type="FareProductCodeType">
+			<xs:element name="FareProductCode" type="FareProductCodeType">
```
Aurige, line 258: not sure why Code over Id ? in NeTEx the FP has an **Id** and a **PrivateCode**, but not a **Code**

ue71603: because it is a FareProductCodeType... we can harmonise it the other way.

ue71603: Transmodel has it as FareProductIdType, so I switch that part.

AndreasAtSBB (dismissed):
Changed text also in the html file: Changed "FareProductCodeType" to "FareProductIdType". OK.

trurlurl (dismissed):

Aurige (dismissed):

skinkie (approved):

sgrossberndt (approved):

Aurige (approved):

PR #290, opened 2022-12-20
Switches for accessibility information

trurlurl:
From the meeting: adding the ability (OJP_TripRequest) to specify the accessibility-related information to return as it may differ depending on the user profile and on what data will be used and presented.


ue71603:
redone directly in VDV repository  (https://github.com/VDVde/OJP/pull/291 ) and therefore closed here

trurlurl:
Removed *documentation* label, documentation will be done for the follow-up PR.

PR #238, opened 2022-08-26
Add AccessFeatureStatus on PathLink

ue71603:
AccessFeatureStatus on a PathLink to indicate a broken access equipment.

https://github.com/VDVde/OJP/issues/194 asks for data to tell whether, e.g., an elevator is broken, but to nevertheless show the route so that the passenger knows that this route would exist but is currently not feasible.

EquipmentStatusEnumeration has been added from NeTEx to indicate the status of an access equipment. The following would indicate that the elevator is broken:

PathLink.AccessFeatureStatus: notAvailable (new)
PathLink.AccessFeatureType: elevator

The broken elevator makes the complete TransferLeg unfeasible - this should be flagged in the TransferLeg (e.g., TransferLeg.Feasibility : accessEquipmentOutOfService) or/and at the level of the Trip.
What solution would you prefer?

To also handle the case that the TranferLeg is not feasible due to a SituationFull (can that happen?), this could be indicated by:
TransferLeg.Feasibility: seeSituations

Replaces #235

PR #291, opened 2022-12-22
Switches for accessibility information

ue71603:
From the meeting: adding the ability (OJP_TripRequest) to specify the accessibility-related information to return as it may differ depending on the user profile and on what data will be used and presented.


AndreasAtSBB (approved):
Is OK.

skinkie:
What I miss in this pull request is the actual 'result structure' for example I want to know what happens if the realtime status would not be available.

trurlurl:
@skinkie To answer your question as I understand it: *IncludeAccessFeatureStatus* (in this PR) is used to request or not real time information about access features. As for the result, see PR #238 - there is *AccessFeatureStatus* (and *AccessFeatureStatusText* and *SituationFullRef*) to provide the real time status. If some data is not available, this can be signalled using *AccessFeatureWithUnknownAvailability*. 

Additionally, there is the problem code *TRIP_REALTIMEINCOMPLETE* that could be used, It should probably also be used by TripRefineRequest, shouldn't it?

skinkie (dismissed):

skinkie (dismissed):

OJP/OJP_JourneySupport.xsd
```diff
@@ -665,6 +675,18 @@
...
 		</xs:sequence>
 	</xs:group>
+	<xs:simpleType name="AccessibilityDetailsProfileEnumeration">
+		<xs:annotation>
+			<xs:documentation>Allowed values for AccessibilityDetails.</xs:documentation>
+		</xs:annotation>
+		<xs:restriction base="xs:string">
+			<xs:enumeration value="visualImpairment"/>
+			<xs:enumeration value="auditoryImpairment"/>
+			<xs:enumeration value="mobilityImpairement"/>
```
sgrossberndt, line 685: mobilityImpairment

sgrossberndt (approved):

skinkie (approved):

PR #273, opened 2022-12-11
Routing for accessibility, without Scores

trurlurl:
This is an alternative proposal to #254, keeping BaseTripMobilityFilterGroup (almost) unchanged but nevertheless allowing for specifying low level accessibility needs and preferences.
 

- AccessibilityPreferences have been introduced to allow for minimizing obstacles like stairs even when there is no route with zero stairs (i.e., when NoStairs=true does not find any result).
- It is possible to do a search for a user with both visual and mobility impairment.

PR #254, opened 2022-10-07
Scores for accessibility routing and other preferences

trurlurl:
This is a proposal addressing #187 in a way that is compatible with the vision / roadmap discussed in #221,

Basic idea: Route plannig tries to optimize *Scores* which can be defined as a weighted sum over lower level *Preferences*.

I think to get a list of Trips the user is really happy with, two runs of a planning algorithm will typically be required (unless we had an algorithm optimising multiple Scores in parallel, perhaps as mentioned by Alexandre?): 
- a run optimizing a "fastest" Score on first priority, and, e.g., a "accessibility" Score on a second priority (to get the traditional fastest Trips with additonal accessiblity optimisation and information);
- and a second run optimizing the other (e.g., "accessibility") Score on first priority (to get Trips that optimally fit the user's special needs, but that may result in Trips normally considered too late or too long).

PR #293, opened 2023-01-06
Routing for accessibility

trurlurl:
This is an alternative proposal to PR #254, keeping BaseTripMobilityFilterGroup (almost) unchanged but nevertheless allowing for specifying low level accessibility needs and preferences. It replaces PR #273 - the code is identical but was previously in another, cloned repository.

- AccessibilityPreferences have been introduced to allow for minimizing obstacles like stairs even when there is no route with zero stairs (i.e., when NoStairs=true does not find any result).
 - It is possible to do a search for a user with both visual and mobility impairment.

The discussion in Köln has been taken into account to my best understanding.


OJP/OJP_Trips.xsd
```diff
@@ -184,23 +189,56 @@
...
 			</xs:enumeration>
-			<xs:enumeration value="earliestArrival"/>
-			<xs:enumeration value="latestDeparture"/>
-			<xs:enumeration value="earliestArrivalAndLatestDeparture"/>
 		</xs:restriction>
 	</xs:simpleType>
 	<xs:complexType name="NotViaStructure">
 		<xs:annotation>
-			<xs:documentation>NNot-via restrictions for a TRIP, i.e. SCHEDULED STOP POINTs or STOP PLACEs that the TRIP is not allowed to pass through</xs:documentation>
+			<xs:documentation>Not-via restrictions for a TRIP, i.e. SCHEDULED STOP POINTs or STOP PLACEs that the TRIP is not allowed to pass through</xs:documentation>
```
JonaVBB, line 241: Not passing through or only not allowed to change there?

OJP/OJP_Trips.xsd
```diff
@@ -129,9 +134,9 @@
 		</xs:annotation>
 		<xs:sequence>
 			<xs:group ref="BaseTripMobilityFilterGroup"/>
-			<xs:element name="LevelEntrance" type="xs:boolean" default="false" minOccurs="0">
+			<xs:element name="LevelEntranceOrBoardingAid" type="xs:boolean" default="false" minOccurs="0">
```
JonaVBB, line 137: In my experience it makes a difference if there is a LevelEntrance or a BoardingAid. A LevelEntrance allows a self-reliant usage of the vehicle while a BoardingAid may involve other people (personnel) to arrange the boarding. Such an assisted boarding is not regarded as a a self-reliant usage of the vehicle.

OJP/OJP_Trips.xsd
```diff
@@ -220,6 +258,23 @@
...
 		</xs:sequence>
 	</xs:complexType>
+	<xs:simpleType name="AccessibilityRoutingPreferenceEnumeration">
+		<xs:annotation>
+			<xs:documentation>Allowed values for a AccessibilityRoutingPreference.</xs:documentation>
+		</xs:annotation>
+		<xs:restriction base="xs:string">
+			<xs:enumeration value="levelEntranceOrBoardingAid"/>
+			<xs:enumeration value="noSingleStep"/>
+			<xs:enumeration value="noStairs"/>
```
JonaVBB, line 268: Just for my understanding: noSingleStep automatically also excludes noStairs?

OJP/OJP_Trips.xsd
trurlurl: @JonaVBB This is a very pertinent remark. It means that LevelEntrance is generally preferred over LevelEntranceOrBoardingAid, and we should be able to take this into account. 

I'd propose to have both LevelEntrance and LevelEntranceOrBoardingAid as strict filter options. (LevelEntranceOrBoardingAid being more important since it allows for finding the routes feasible for an unassisted wheelchair)

For the AccessibilityPreferences, we need to add levelEntrance. (levelEntronceOrBoardingAid may still be useful, but less important)

If you agree, I'll make these changes.

OJP/OJP_Trips.xsd
trurlurl: @JonaVBB It only excludes single steps, otherwise it should rather be called *noStep* - which would probably as well include *levelEntrance*

ue71603:
@trurlurl  The PR is fine for me. My suggestion: You clear the open points with @JonaVBB  and then I create it directly from changes_1.1 for the final approval.

ue71603 (approved):

OJP/OJP_Trips.xsd
JonaVBB: Totally agree.

OJP/OJP_Trips.xsd
JonaVBB: It seems to me that we mean the same thing. Maybe my question was misleadingly formulated. I meant: if a user requests noSingleStep I understand that the user is not able to overcome a single step. So, I would furthermore assume that this user is then also not able to overcome stairs. Right?
I agree that noStep would include levelEntrance.

trurlurl: @JonaVBB In the case of ***N**oSinlgeStep* as filter option (hard constraint) you may be right (though I'm not sure, it might be more precise to interpret it by just excluding the stated equipment, i.e., single steps).

In the case of the preferences (*noSingleStep* is from AccessibilityRoutingPreferenceEnumeration), my short answer would be: *noSingleStep* is only about single steps. The user does not *request* noSingleStep. He just prefers minimizing the number of single steps; minimizing stairs would have to be done separately (so as not to mix apples and pears when counting in order to minimize) by additionally specifying *noStairs*. In PR #273 I tried to give an example how to use the preferences. Do you have an idea whether & how to make this more clear for future users?

skinkie (dismissed):

ue71603 (dismissed):

sgrossberndt:
Merged branch 'changes_for_v1.1' into routing_for_accessibility to make it mergeable.

ue71603 (dismissed):

OJP/OJP_Trips.xsd
```diff
@@ -108,7 +108,12 @@
 			</xs:element>
 			<xs:element name="OptimisationMethod" type="OptimisationMethodEnumeration" minOccurs="0">
 				<xs:annotation>
-					<xs:documentation>the types of algorithm that can be used for planning a journey (fastest, least walking, etc)</xs:documentation>
+					<xs:documentation>The types of algorithm that can be used for planning a journey (fastest, least walking, etc)</xs:documentation>
```
skinkie, line 111:
```suggestion
					<xs:documentation>The types of algorithm that can be used for planning a journey (fastest, least walking, etc.)</xs:documentation>
```

OJP/OJP_Trips.xsd
```diff
@@ -108,7 +108,12 @@
...
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="AccessibilityRoutingPreference" type="AccessibilityRoutingPreferenceEnumeration" minOccurs="0" maxOccurs="unbounded">
+				<xs:annotation>
+					<xs:documentation>If bestAccessibility is chosen as OptimisationMethod, the algorithm tries to minimise obstacles and sections without accessibility aids. The user may specify the criteria to be considered. The order of priority is given by the sequence in which they are listed (starting with the highest priority). Higher priority means, in terms of a cost function, that a change at a higher priority can never be compensated by changes at the lower priorities.</xs:documentation>
```
skinkie, line 116: This is a very interesting statement. I would like to discuss it.

OJP/OJP_Trips.xsd
```diff
@@ -921,7 +986,12 @@
 			</xs:element>
 			<xs:element name="OptimisationMethod" type="OptimisationMethodEnumeration" minOccurs="0">
 				<xs:annotation>
-					<xs:documentation>the types of algorithm that can be used for planning a journey (fastest, least walking, etc)</xs:documentation>
+					<xs:documentation>The types of algorithm that can be used for planning a journey (fastest, least walking, etc)</xs:documentation>
```
skinkie, line 989:
```suggestion
					<xs:documentation>The types of algorithm that can be used for planning a journey (fastest, least walking, etc.)</xs:documentation>
```

sgrossberndt:
@skinkie I integrated those etc typo fixes already in https://github.com/VDVde/OJP/commit/a96652093b00df6ed2dbcad41c65da16eec7f023

ue71603:
André: Jona wants to check it again.

OJP/OJP_Trips.xsd
JonaVBB: Same for me.

OJP/OJP_Trips.xsd
JonaVBB: @trurlurl OK, got it. Thank you.
Maybe it helps to emphazise in the documentation that noSingleStep and noStairs are technically independent from each other, although their wording might lead to interpretation that noSingleStep "includes" noStairs.
How do you distinguish between _minimizing_ steps/stairs and _totally avoiding_ them?

trurlurl:
I added *AccessibilityRoutingWeight* in the latest commit. This is the improvement mentioned towards the end of section 5.4.1 (*Multiple optimisation methods*). The PR now contains the complete solution as discussed in section 5.4.1. I hope this makes things sufficiently explicit and clear.

Note that the planning algorithm of a system may not support such a more general cost function defined by the *AccessibilityRoutingWeight*s.  This is not a problem, in that case it could fall back to the simpler solution without the weights.

The idea is to make the interface sufficiently general to be ready for more sophisticated algorithms that may be used in the future.

ue71603:
@normanoffel @herlitze I disucssed with André that he should provide the concrete example. It is now available in tihs commit. During the next meeting I suggest we schedule a meeting for a discussion with whoever wants to participate to settle this one. I hope this is ok.

herlitze:
@ue71603 Yes, I think we need this discussion.

ue71603:
From Malte to @trurlurl 
today I had an internal discussion about accessibility routing and found real world examples to be helpful for the discussion.
Here are two examples for accessibility routing.

"Normal" vs. stepfree routing:
image

"Normal" vs. limited vision routing:
image
Could you please provide examples as the above for the meeting next week so we have specific use cases to find solutions for?

ue71603:
* Remove multiple criteria
* Additional criteria for accessibility, but only single ones, with LeastNumberOfSteps, ....
* Describe behaviour that is expected.
* Declare which criteria that are core
*

ue71603:
see https://github.com/VDVde/OJP/pull/353

ue71603:
will be done here: https://github.com/VDVde/OJP/pull/353  some documentations points may be copied to the specifciation

PR #294, opened 2023-01-11
Obsolete code in OJP_RequestSupport.xsd

trurlurl:
The removed elements don't seem to be used anywhere in the XSD. Is it really o.k. to remove them?


ue71603 (approved):

Aurige (approved):

skinkie (changes requested):
I would not want to use the SubscriptionRequest. Does the subscription request itself exist with the removal of the SubscriptionRequestContextGroup?

ue71603:
Norman: Does this come from TripMonitoring?
Stephan: Guess it was.
So we can remove it now. If we do TripMonitoring we can do it.
Stefan: Concur. I want TripMonitoring

sgrossberndt (approved):

Aurige (approved):

skinkie (approved):

skinkie (approved):

sgrossberndt (approved):

PR #297, opened 2023-01-17
Create OJP.spp

Aurige:
Addition XML Spy Project
Very convenient for XML Spy users !


ue71603:
I add it to be discussed: We had it at the beginning and removed it about 2 years ago.

skinkie:
Bigger question: who maintains it, and can it be automatically generated?

PR #298, opened 2023-01-17
Update trip refinement

ue71603:
Based on Malte's comment:  Isn't there still a parameter missing in the TripRefine request to select a specific sharing provider when requesting a specific trip with that provider?

And some discussions I had with Andreas and André.

It is now possible to control more, what refinement should do, which system should do it.
A lot will need to be described in the specification.


skinkie (approved):

OJP/OJP_Trips.xsd
```diff
@@ -1062,6 +1062,18 @@
 			</xs:element>
 		</xs:sequence>
 	</xs:group>
+	<xs:group name="TripRefinementDataFilterGroup">
+		<xs:annotation>
+			<xs:documentation>Request data filters for Refinement</xs:documentation>
+		</xs:annotation>
+		<xs:sequence>
```
herlitze, line 1069: In the workflow we discussed, we only had a single operator passed in the refinement. Do you have an usecase when a sequence is needed?

OJP/OJP_Trips.xsd
```diff
@@ -1072,9 +1084,30 @@
...
+			</xs:element>
+			<xs:group ref="TripRefinementDataFilterGroup"/>
+			<xs:element name="KeepLegs" type="xs:boolean" minOccurs="0" default="true">
+				<xs:annotation>
+					<xs:documentation>Shall the legs be kept or can neighbours be recalculated in one go and changed.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="RefitLegs" type="xs:boolean" minOccurs="0" default="false">
+				<xs:annotation>
+					<xs:documentation>With refit, the leg is/legs are recalculated from the end point of the adjascent legs. If necessary new foot paths are added.</xs:documentation>
```
herlitze, line 1100: I don't understand the difference between KeepLegs and RefitLegs, could they be combined?

OJP/OJP_Trips.xsd
```diff
@@ -1072,9 +1084,30 @@
...
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="CheckAvailability" type="xs:boolean" minOccurs="0" default="false">
+				<xs:annotation>
+					<xs:documentation>Checks the availability of relevant legs depending on the mode/submode. Operators or system references must be set to do this (uses OJPAvailabilityRequest indirectly)</xs:documentation>
```
herlitze, line 1105: Let us talk about the workflow you are expecting behind this on Friday.

OJP/OJP_Trips.xsd
skinkie: It seems to be that with taxi, multiple companies may be choosen from.

OJP/OJP_Trips.xsd
ue71603: to be discussed label already added

OJP/OJP_Trips.xsd
ue71603: Matthias: There might exist aggregators. 
Malte: One trip result by operators? Currently a general taxi and  sharing, then update in the refinement. So I thought that refinement go to one operator. That's what expected. 
==> first do the chapter in the specification, discuss it, then finalize this PR (chapter 7.9)

OJP/OJP_Trips.xsd
ue71603: Norman: I expect that the responsive system  to refit things. 
Malte: Shouldn't we go for an enum what to do. not two boolean.s

normanoffel:
1. It is probably a good idea to indicate in the TripResponse which TripLeg needs to be refined because it is only an estimate.
2. I assume the SystemID meant to address responding systems which could be added to the documentation annotation.
3. What is the purpose of the SystemID when for example compared to the OperatorRef? Isn't the original responding system always the one to be asked for refinement as well?

ue71603:
Use case 5.24 is adapted.

7.9 and 8.10 will have to be updated after this PR is done. I will add some more information there while finalising this PR.

ue71603:
simpliefied a lot. So I guess we have no longer issues with it?  @normanoffel  @herlitze 

@AndreasAtSBB : Pls check, if this still fits the bill for your document.

examples/functions/RefinementRequest.xml
```diff
@@ -0,0 +1,153 @@
...
+	<OJPRequest>
+		<ServiceRequest>
+			<RequestTimestamp>2022-12-19T12:00:00Z</RequestTimestamp>
+			<RequestorRef>IRMA</RequestorRef>
+			<ojp:OJPTripRefineRequest>
+				<RequestTimestamp>2022-12-19T12:00:00Z</RequestTimestamp>
+				<MessageIdentifier>4711</MessageIdentifier>
+				<ojp:RefineParams>
+					<ojp:RefineLegRef>1</ojp:RefineLegRef>
+					<ojp:RefineLegRef>2</ojp:RefineLegRef>
```
herlitze, line 12: There is only one leg passed, but refinement of leg 1,2,4 is requested.

examples/functions/RefinementResponse.xml
```diff
@@ -0,0 +1,111 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
herlitze, line 1: What was refined, I don't see the difference to the passed trip.

ue71603: @herlitze  This was only a test for elements and structure not content. But you are right we should provide a "good" example.

ue71603: now refinement done

AndreasAtSBB (dismissed):

OJP/OJP_JourneySupport.xsd
```diff
@@ -243,9 +251,9 @@
 			<xs:element name="Extension" type="xs:anyType" minOccurs="0"/>
 		</xs:sequence>
 	</xs:complexType>
-	<xs:complexType name="TripPlaceStructure">
+	<xs:complexType name="TripLocationStructure">
 		<xs:annotation>
-			<xs:documentation>A trip place represents the current logical position of a  journey service. It can be used similarly to a place e.g. for starting a new trip requests from within this service. A trip place does not(!) describe the relative position of a traveller within a vehicle, e.g. the seat.</xs:documentation>
+			<xs:documentation>A trip location represents the current logical position of a  journey service. It can be used similarly to a place e.g. for starting a new trip requests from within this service. A trip location does not(!) describe the relative position of a traveller within a vehicle, e.g. the seat.</xs:documentation>
```
trurlurl, line 256:
```suggestion
			<xs:documentation>A trip location represents the current logical position of a  journey service. It can be used similarly to a place e.g. for starting a new trip request from within this service. A trip location does not(!) describe the relative position of a traveller within a vehicle, e.g. the seat.</xs:documentation>
```

trurlurl: Perhaps I miss something - but I don't see how the three elements in the sequence could indicate "the current logical position of a journey service".  I would rather expect elements for current or next call, or vehicle position to do that.

OJP/OJP_JourneySupport.xsd
```diff
@@ -502,21 +514,21 @@
 						<xs:documentation>Static place.</xs:documentation>
 					</xs:annotation>
 				</xs:element>
-				<xs:element name="TripPlace" type="TripPlaceStructure">
+				<xs:element name="TripLocation" type="TripLocationStructure">
 					<xs:annotation>
-						<xs:documentation>Place within a (moving) vehicle.</xs:documentation>
+						<xs:documentation>Location within a (moving) vehicle.</xs:documentation>
```
trurlurl, line 519: Isn't that in direct contradiction to the description of *TripLocationStructure*, line 256?

trurlurl (changes requested):

skinkie (approved):

OJP/OJP_Trips.xsd
```diff
@@ -1472,16 +1484,22 @@
...
-			<xs:element name="RefineLegRef" type="xs:NMTOKEN" maxOccurs="unbounded">
+			<xs:element name="RefineLegRef" type="xs:NMTOKEN" minOccurs="0" maxOccurs="unbounded">
+				<xs:annotation>
+					<xs:documentation>Refers to the legs to be refined by the server. If none is given, then all legs are open for refinement (depending if the relevant system can refine them).</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="SystemId" type="xs:normalizedString" minOccurs="0">
 				<xs:annotation>
-					<xs:documentation>Refers to a leg to be refined by the server.</xs:documentation>
+					<xs:documentation>System reference to use for the refinement. If not specified the origin system of each leg are used for the refinement.</xs:documentation>
```
sgrossberndt, line 1494: > If not specified the origin system of each leg are used for the refinement
Should this be "the origin systems"?

ue71603, line 1494:
```suggestion
					<xs:documentation>System reference to use for the refinement. If not specified the origin systems of each leg are used for the refinement.</xs:documentation>
```

skinkie (dismissed):

skinkie (approved):

trurlurl (approved):
What irritates me a little: If the idea is that TripRefinement should not rely on trip recalculation - don't changed parameter values in *OperatorFilter* and *IncludeAlternativeOptions* require invoking the trip planner?

skinkie (approved):

PR #303, opened 2023-01-20
Update OJP_Trips.xsd

ue71603:
Malte has some typos found, but it was already merged. His comments are now in this PR

### DIFF
--- a/OJP/OJP_Trips.xsd
+++ b/OJP/OJP_Trips.xsd
@@ -791,7 +791,7 @@
 			</xs:element>
 			<xs:element name="TurnAction" type="TurnActionEnumeration" minOccurs="0">
 				<xs:annotation>
-					<xs:documentation>The range of alternative turns that can be described.</xs:documentation>
+					<xs:documentation>The range of possible turns that can be described.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
 			<xs:element name="RoadName" type="InternationalTextStructure" minOccurs="0">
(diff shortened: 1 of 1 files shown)


skinkie (approved):

AndreasAtSBB (approved):
Changes OK.

skinkie (changes requested):
Want to include:
https://github.com/VDVde/OJP/pull/238#pullrequestreview-1263383977

skinkie (approved):

OJP/OJP_Trips.xsd
```diff
@@ -943,13 +940,16 @@
 					<xs:documentation>If there are more than 2 lanes, then TurnAction half_right,right,sharp_right may help decide</xs:documentation>
 				</xs:annotation>
 			</xs:enumeration>
-			<!-- undefined values -->
 			<xs:enumeration value="unknown_turning_kind"/>
+			<!-- additional enum values for individual vehicles -->
```
sgrossberndt, line 944: Did you move those comments on purpose?

ue71603: no, I think this is something that the post processing in the script does. It doesn't like  style comments within xs:enumeration. Perhaps we should remove them.

herlitze (approved):
Thanks!

sgrossberndt:
@skinkie Why did you merge this? The problem of those comments has not been solved. Either they should be put back to where they were or removed - but not be in the bottom of the element.

PR #313, opened 2023-01-24
First fare examples

ue71603:
First fare examples.. .I needed those for our OJPFare2TOMP mapping


skinkie (approved):

trurlurl (approved):
Is or should this be a more or less realistic example? FareAuthorityFilter filters for NOVA, but there are two FareResults linked to two fare authorities, NOVA and Mybuxy.

PR #314, opened 2023-01-24
Path to xsd corrected in xml examples

ue71603:
After the reorgnanisation the relative paths to the xsd in the xml were incorrect. This is now corrected.


skinkie (commented):
I think this is still not 'correct'.

ue71603:
> I think this is still not 'correct'.

they validate in XML Spy. What makes you unhappy?

skinkie:
What do you think happens when we migrate to the upstream SIRI branch...

ue71603:
nothing hopefully. It will link to the same place in the local copy, I guess.

AndreasAtSBB (approved):
File path fixed.

skinkie (approved):

PR #316, opened 2023-01-28
Coherent use of ObjectIdType

trurlurl:
Some Result, Trip, and Leg IDs had been changed from xs:NMTOKEN to ObjectIdType, see #215.
References have not been consistently adapted accordingly. 

I hope it is correct that only some of the Results have an id (FareResult, StopEventResult, TripResult, and MultiPointTripResult).


skinkie (changes requested):
lets discuss if we would like to have an TripIdType, LegIdType, etc.

trurlurl:
I'm adding the discussion label, to discuss whether we'd like to have not simply ObjectIdType (for NeTEx objects), but TripIdType, LegIdType, etc. 

Would that be wanted mainly for validation purposes? What kind of validation tool? If it were something like "Data4PT NeTEx Validator", then I'd imagine that one wanted to do more intelligent checks than just syntax-based type checking - e.g., checking whether the referenced leg id really exists somewhere. But in order specify more intelligent checks, one would probably look at each element separately and not rely on the type.

ue71603:
Stefan: For validation purposes this might be a good idea.
Norman: Why do we need identity constraint?
Stefan: From code generation it would help as well.

Appreciated, but not pressing.
==> if we find somebody to do it.

skinkie (approved):
Improvement in the future.

ue71603 (dismissed):

ue71603 (approved):

skinkie (approved):

ue71603 (approved):

PR #324, opened 2023-03-02
Adding a parameter in FareResult that tells, if the result was estimated

ue71603:
In many cases the FareResult is only an estimate. We should be able to communicate this.


trurlurl (commented):
The PR wants to merge into *master*...?

ue71603:
> The PR wants to merge into _master_...?

now corrected.

trurlurl (dismissed):

OJP/OJP_Fare.xsd
```diff
@@ -124,6 +124,11 @@
 					<xs:documentation>Problems related to this FARE result.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element name="FareEstimated" type="xs:boolean" default="true" minOccurs="0">
+				<xs:annotation>
+					<xs:documentation>Is set to true, when the fare is only estimated by the service. If not set, it should be assumed to be true.</xs:documentation>
```
sgrossberndt, line 129: 1. Proposed documentation: "Whether the fare is only estimated by the service. Default is TRUE."
2. I wonder whether setting default as true is sensible, isn't the "usual" case rather it is calculated? What does "estimated" mean, how to react to this as a receiver?

skinkie (approved):

sgrossberndt (changes requested):

OJP/OJP_Fare.xsd
ue71603, line 129:
```suggestion
					<xs:documentation>Whether the fare is only estimated by the service. Default is false.</xs:documentation>
```

OJP/OJP_Fare.xsd
```diff
@@ -124,6 +124,11 @@
 					<xs:documentation>Problems related to this FARE result.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element name="FareEstimated" type="xs:boolean" default="true" minOccurs="0">
```
ue71603, line 127:
```suggestion
			<xs:element name="FareEstimated" type="xs:boolean" default="false" minOccurs="0">
```

sgrossberndt:
What does "estimated" mean, how to react to this as a receiver?

sgrossberndt (dismissed):

ue71603:
@sgrossberndt  Improve documentation in XSD or in specification.

The idea is: "Estimated can mean different things: The price may vary e.g. depending on age, PassangerCategory, TravelClass, EntitlementProducts). It may also mean, that the fare structure that the fare service uses is not really accurate or that other factors like time a shared car is used, may play into the price. Usually a client should then should present the price as approximate and probably lower limit of the real price."

trurlurl (approved):

skinkie (dismissed):

sgrossberndt (approved):

skinkie (approved):

PR #326, opened 2023-03-02
IsAlternativeOption

ue71603:
To make it clear, if the trip is the result from alternative options checking.


trurlurl (approved):

skinkie:
Shouldn't it then have a reference to the alternative? If it already has it, why explicitly mention it?

ue71603:
It doesn't have a reference and it is not clear, why the server did it. The server may have a myriad ideas how to define the alternative. So it just mentions it, that the result may not be considered optimal by what the request asked.

trurlurl (dismissed):

ue71603:
leave it at that.

trurlurl (approved):

ue71603:
@skinkie : ready to merge....

PR #330, opened 2023-03-07
copied SIRI 2.1 and adapted imports

ue71603:
without copying from upstream to see, if it works....


skinkie (approved):
I have checked it with XMLlint, works.

skinkie (approved):

trurlurl (changes requested):
I don't know whether it makes sense to complain about minor problems in external code...

AndreasAtSBB (approved):
Changes in html dok seams to be fine.

trurlurl:
@ue71603 Does this need some update to the documentation (the *documentation* labe is set)?

ue71603:
@trurlurl 
* If it is mentioned somewhere in the documentation which SIRI is used, it should be adapted.
* If we use SIRI tables in chapter 8 perhaps a brief check is necessary, if it still fits. 
Otherwise I think that the real changes will manifest in other PR and the documentation comes from there.

PR #329, opened 2023-03-04
Use SIRI v2.1 from upstream as git submodule

skinkie:
At least try if this would work.

PR #331, opened 2023-03-07
Use SIRI 2.0q from git submodule

sgrossberndt:
Replace local siri files by SIRI 2.0q as git submodule

git rm ./siri
git submodule add -b 2.0q --name SIRI https://github.com/SIRI-CEN/SIRI ./siri


sgrossberndt:
Using upstream 2.0q fails due to `siri/xsd/xml/xml.xsd` which uses languages defined with uppercase letters while our examples use lowercase ones. an updated xml.xsd exists which also allows lowercase ones (but is not in the siri 2.0q branch)

skinkie:
Already implemented in #329?

ue71603:
I can't accept the PR for the time being. It breaks my ability to work on the repository:
image

Afterwards I have to reclone the repository. This is bad enough in a single branch. But if changes_v1.1 does this, I will not be able to work on the PR any longer until this is fixed. And this is for me optional work for the time being. I can work with a copied SIRI 2.1 for the work until we hand the specificiation to CEN and then I can think about why git for desktop on Windows doesn't behave as it should

* It seems that this is a known bug: https://github.com/desktop/desktop/issues/13016 
* https://github.com/desktop/desktop/issues/13016 (this worked temporarily but as you only have a git reference to SIRI, it will no longer validate anything on my machine.
* https://lore.kernel.org/git/7C3FF354-F10E-4822-ACF8-31417B05C099@gmail.com/t/

Deactivating submodules will result in an empty SIRI folder and then validation might be a little bit tricky:

sgrossberndt:
> Already implemented in #329?

While you were working on v2.1 I wanted to test first whether v2.0q works with out setup. Since v2.1 works now and this one does not it is obsolete.

@ue71603 Did you try the workaround [from the issue](https://github.com/desktop/desktop/issues/13016)? 

> The workaround for this issue is to remove the following lines from .git/config before attempting a branch checkout:
> 
> ```
> [submodule]
>     active = .
> ```

PR #333, opened 2023-03-09
Aligning MultiPointTripPolicyGroup with TripPolicyGroup

trurlurl:
Shouldn't the recent additions to the TripParams also be available to MultiPointTripRequest? 

*OptimisationMethods*, *ItModesToCover*, *ConsiderElevationData*


ue71603 (approved):

skinkie (approved):

sgrossberndt (changes requested):
There is a BaseTripPolicyGroup which is used by `MultiPointTripPolicyGroup` and `TripPolicyGroup` only. In my opinion all elements belonging to both MultiPointTrip and Trip should be put into BaseTripPolicyGroup. With the current state of this change the OptimisationMethod-choice would be aligned but at least `ConsiderElevationData` makes sense in `MultiPointTripPolicyGroup` as well, right? With the current definition of `ItModesToCover` in my opinion it does not make sense to add it to `MultiPointTripPolicyGroup` as well, wich would mean all elements are part of `BaseTripPolicyGroup`, only `TripPolicyGroup` has the additional `ItModesToCover` and only `MultiPointTripPolicyGroup` has the additional `MultiPointType`.
`ConsiderElevationData` must be moved in front of `ItModesToCover` in order to have it in `BaseTripPolicyGroup`, but it was only recently added, so its no big deal, right?

ue71603:
@trurlurl Can you do all the changes according to Stephan or should I do it?  And yes: They should be harmonized fully.

sgrossberndt:
Working on it.

sgrossberndt (approved):
will update again

sgrossberndt:
`ItModesToCover` has the definition:

> For each mode in this list a separate monomodal trip shall be found - in addition to inter-modal solutions.

Does this make sense for a `MultiPointTripPolicyGroup` for you? I'd rather not add it there, I think it is on purpose this was only part of `TripPolicyGroup` and not `MultiPointTripPolicyGroup`.

trurlurl:
> `ItModesToCover` has the definition:
> 
> > For each mode in this list a separate monomodal trip shall be found - in addition to inter-modal solutions.
> 
> Does this make sense for a `MultiPointTripPolicyGroup` for you? I'd rather not add it there, I think it is on purpose this was only part of `TripPolicyGroup` and not `MultiPointTripPolicyGroup`.

I could imagine that it could make sense when distributing a search for monomodal trips by calling OJPMultiPointTripRequest for parts of the trip - and that these parts should also be monomodal.

ue71603:
@sgrossberndt  I think you are right about ItModesToCover. That shouldn't be done in MPTR

ue71603:
I think we can define MPTR that it is used in the middle of distributed trip planning and I think if one ones monomdal trips through Europe, one does this directly on one iv trip planner.

ue71603 (approved):

skinkie (approved):

Issue #305, opened 2023-01-20
Study if TripMonitoring should be added (from TRIAS)


ue71603:

ue71603:
Malte: a lot of work without the distributed context. Use case is needed and what behaviour is necessary.
You calculate a trip and afterwards push notifications. It can be quite hard in the distributed context.
Norman: Not trivial. In our research project we used it, but only asked a responding system. The central system then knew about the whole trip. Then it can work on that. we have some experience in that regard. Trip Monitoring is not part yet.  We would like to have it. 
Stefan: What is not trivial? distribution or monitoring? 
Norman: You have a TripResponse with different parts from different system. The trip monitoring needs to monitor the whole trip somehow. So it would need to be done in the distributing system.
Stefan: The problem I think it is the subscriptions right. Information with hysteresis . When to notify.  Part of the scope of the solution.
Norman: We were able to use TripMonitoring as it is.  We had a central component. Separate work stream. 
==> its own meeting.
Matthias: Which part.
Malte: Analysis of threshold not in TripInformationRequest. What woudl be the expected behaviour. Does it has the trip and recalculates it and checks it? Or does it give just back the things and analysis is done by the client.
Malte: Trias did the analysis, then we need a new request. With TIR the logic would be on the client side.
Stefan: The unsubscription might be a problem. Is this solved in TRIAS.  
Norman: I would have to revisit it.  Part of Trias. If the trip is done, the subscription is cancelled anyhow.
Stefan: SIRI ET or this request?
Christophe: More events in SIRI ET and don't forget about SIRI SX. Regarding only ET may not do the trick.
Claus: Would the trip monitoring, monitor the complete trip or only individual legs. In the second case it can be done with SIRI subscription with one trip. If it is mixed from different sources, then we need both.
Norman: You had a TripResponse, one could do trip leg monitoring or one has a distributing system that is central and can do the whole trip. Please do that for me.
Stefan: From client this is nice (trip monitoring). 
Christophe: Not really for end users.
Matthias: Also perhaps special focus on transfers. 

Result: New Service is added.

Involved are:
- Stefan,
- André
- Chistophe
- Norman
- Malte

Next steps:
* First PR based on TRIAS
* chapters prepared in the spec
* Meeting with interested people.

ue71603:
Dirk: Ending subscriptions with XSD InitialTerminationTime is mandatory.

PR #341, opened 2023-03-21
Trip monitoring service

trurlurl:
PR for issue #305. First draft, to be discussed and adapted.

We need to write in the documentation, that it only works for  interacting between trusting systems as denial of service attacks are possible.


trurlurl:
@sgrossberndt Could you please help with identifying what conventions the two groups violate? 

Run ./docs/validate-schema-conventions.sh
Validating OJP Schema conventions ...
Skipping... OJP.xsd
Checking... OJP_Requests.xsd
[ERROR] Group could not be resolved: AlertSettingsGroup
[ERROR] Group could not be resolved: TripMonitoringResponseGroup

OJP/OJP_TripMonitoring.xsd
```diff
@@ -0,0 +1,192 @@
...
+				<xs:annotation>
+					<xs:documentation>Response context aggregated for all TripInfoResults.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="AlternativeTrip" type="TripResultStructure" minOccurs="0" maxOccurs="unbounded">
+				<xs:annotation>
+					<xs:documentation>Alternatives to a delayed or otherwise compromised trip.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="FacilityCondition" type="siri:FacilityConditionStructure" minOccurs="0" maxOccurs="unbounded">
```
ue71603, line 107: no. Not the FacilityCondition in this service. Separate if necessary.

ue71603:
Malte, Stefan and Norman will say how the pub/sub should work (from security).
Stefan: What does Dirk think?
Dirk: Making subscription for others (DoS attacks) might be a problem. Fixing was hard.  Data broker. Discussion in SIRI necessary. Pub/Sub.  
Norman: We should only do payload, if possible in OJP.
Claus: MQTT is better organised. 
Dirk: Topic based architecture will be described. I tried to create rules there a based topic architecture. 
Norman: How does this address the security concern? The only solution is on the transport protocol, not hear. 
Claus: We stick within the SIRI with request/response. Outside pub/sub outside.

ue71603:
@herlitze @skinkie @normanoffel @sgrossberndt 
Possible security improvement:
The problem arises from the fact that
`			<DelegatorAddress>https://182.82.1.23:89912/endpoint</DelegatorAddress>
			<DelegatorRef>23188</DelegatorRef>
`
are not controlled. So it is either a completly trusted environment between the subscriber and the server or 
DelegatorAddress and DelegatorRef are to compared to saved values from the authentication method (probably in an API manager or proxy).
This could be done by using the id from The authentication in DelegatorRef and looking up the DelegatorAddress.
Or in the API manager DelegatorAddress and DelegatorRef are supressed and replaced by the values stored there. 

Also important might be RequestorRef (is mandatory). This also could be filled in by the API manager (or proxy).

Do you agree that we add something to that effect to the specification (or even the XSD).

OJP/OJP_TripMonitoring.xsd
ue71603: It is working fine. One can name the relevant Facility.
Equipment not here.

OJP/OJP_Requests.xsd
```diff
@@ -299,4 +299,39 @@
...
+					<xs:group ref="AlertSettingsGroup"/>
+					<xs:element ref="siri:Extensions" minOccurs="0"/>
+				</xs:sequence>
+			</xs:extension>
+		</xs:complexContent>
+	</xs:complexType>
+	<xs:element name="OJPTerminateSubscriptionRequest" type="siri:TerminateSubscriptionRequestStructure">
+		<xs:annotation>
+			<xs:documentation>Request element for terminating subscriptions. To change parameters of a subscription, it is necessary to terminate and to re-subscribe. Successful termination is acknowledged by problem code TRIPMONITORING_SUBSCRIPTION_TERMINATED in a delivery.</xs:documentation>
+		</xs:annotation>
```
trurlurl, line 322: According to the examples below, there is no *OJPTripMonitoringDelivery* in response to *siri:SubscriptionRequest* and *siri:TerminateSubscriptionRequest* - and therefore no possibility to directly return any of the *TRIPMONITORING_...* error codes. Is this a problem that needs to be solved or will the first reply (*siri:SubscriptionResponse*, e.g.) be followed by a second reply with a *OJPTripMonitoringDelivery*?

ue71603: Norman: No feedback yet. Investigating further
Matthias: Will check and if necessary error handling to be added.
Security considerations in short best practice paper (for all subscription related things from SIRI)
Andre: Currently we return a TripInformationDelivery on a Journey. We should be able to inform about a transfer leg. the question was, to how to do that.
Matthias: AlternativeTrip
Norman: Would solve it the way it was solved in TRIAS.

* Question to be prepared (Matthias, André)
* Norman asks TRIAS people (at Hafas)
* Adaption of the PR: (a) error handling (b) depending on result of TRIAS inquiry.
* Paper for security considerations added

ue71603:
Malte: Werner: The TripResult only says only hey you need to send a TripInfoRequest.
Matthias: We could already provide an AlternativeTrip.
Malte: The intention was about triggering TripInfoRequest. TRIAS might be flawed there.

TripMonitoring: Trip is broken
-> TripInformationRequests
-> Think about what one gets.
-> TransferLeg needs to be recalculated with TripRequest (with I am not this Vehicle and want to continue to the end)

OJP/OJP_Requests.xsd
ue71603: Norman: We had a research project in Shift2Rail. No production use for TripMonitoring. I can only describe how it was used. Only got notification. One has to add requests.  For more details TripInfoRequest is needed.
Currently no real world use, it might be not a good idea to add it. Perhaps postpone it to a later version. 

Postpone.

sgrossberndt:
Closing this pull request as I am going to delete the changes_for_v1.1 branch. The branch trip_monitoring should be rebased on develop and have a new pull request.

sgrossberndt:
Deleted branch trip_monitoring in favor of https://github.com/VDVde/OJP/tree/feature/trip-monitoring which was rebased on develop.

PR #342, opened 2023-03-21
ITModesToCover -> ModesToCover

ue71603:
also change in behaviour. This would help in having a special mode within. E.g. especially "air" and  "ferry", boat


ue71603:
What do you mean with "this" one of the zillion mode enumerations? Because they don't fit fully. And you know how the MODE discussions is starting to unbound itself from any boundaries.... But I am open to suggestions.

skinkie:
My fundamental problem is here that we start a new enumeration that will absolutely require mapping in the serving application. Rather than passing forward the value of the upstream enumeration.

ue71603:
I agree, but can of worms..... Do you try to do the PR?

ue71603:
We will not do that. Added a new section 5.4.2 telling, that it is the enquirer's work.

PR #348, opened 2023-04-04
adaption of IdTypes

ue71603:
Each Id has now its own type


ue71603:
> I think that at least the following id's would also have to be updated: _FromLegIdRef_, _ToLegIdRef_ (OJP_FareSupport.xsd and OJP_Fare.xsd), _TripId_ (in _TripSummarStructure_), _FromTripIdRef_ and _ToTripIdRef_ (OJP_Fare.xsd). But I somehow don't understand the logic or intent of the _...IdRef..._ names.

Many ref point to a reference structure. But here, when it is just within the Trip I reference directly the ObjectIdType. I adapted it just as you suggested and fixed one that I missed before.

OJP/OJP_Fare.xsd
```diff
@@ -208,22 +214,22 @@
 		</xs:annotation>
 		<xs:sequence>
 			<xs:element ref="FareProductRef"/>
-			<xs:element name="FromTripIdRef" type="ObjectIdType">
+			<xs:element name="FromTripIdRef" type="LegObjectIdType">
```
Aurige, line 217: Shouldn't it be TripObjectIdType ?

ue71603, line 217:
```suggestion
			<xs:element name="FromTripIdRef" type="TripObjectIdType">
```

ue71603: and now it is. tnx

Aurige (approved):

trurlurl (approved):

PR #353, opened 2023-04-13
More optimisation methods, reorganising filters and policy for Trip Request according to discussions

ue71603:
After discussion 2023-04-13. Replaces: https://github.com/VDVde/OJP/pull/293/files#

PR #368, opened 2023-06-10
Relocate BikeTransport

trurlurl:
Problem: *BikeTransport* vanished from *MultiPointTripParamStructure* and from *BookingUserStructure*. To blame is a change done in PR #353. I propose simply reverting that change - i.e., putting *BikeTransport* back to *TripMobilityFilterGroup*. Other solutions would be possible.


ue71603:
@trurlurl  I think it is moved to BaseTripPolicyGroup.  So I think it is still there. Or do you not find it?

trurlurl:
> @trurlurl I think it is moved to BaseTripPolicyGroup. So I think it is still there. Or do you not find it?

@ue71603 No, I don't find it in *BaseTripPolicyGroup*. Even if it were there - it would't appear in *BookingUserStructure* in that case, so that wouldn't be what is wanted, I guess? 

*BaseTripMobilityFilterGroup* would work in that respect, as does the current proposal.

ue71603 (dismissed):

skinkie (approved):

ue71603:
@skinkie : Did we make the thing redudant?

PR #370, opened 2023-06-14
Multiple BookingNotes

ue71603:


trurlurl (approved):

ue71603:
Either one or a new element with multiple elements and Choice

OJP/OJP_FareSupport.xsd
```diff
@@ -446,11 +460,18 @@
...
-			</xs:element>
+			<xs:choice>
+				<xs:element name="BookingNote" type="BookingNoteType" minOccurs="0">
+					<xs:annotation>
+						<xs:documentation>Note about booking the (FLEXIBLE) LINE (depreciated).</xs:documentation>
+					</xs:annotation>
+				</xs:element>
+				<xs:element name="BookingNotes" type="BookingNoteStructure" minOccurs="0">
+					<xs:annotation>
+						<xs:documentation>Notes about booking the (FLEXIBLE) LINE.</xs:documentation>
```
trurlurl, line 471: This sounds like that the notes could only be used for a flexible line. I don't understand why this should be, but I would hesitate using them for any line.

trurlurl (changes requested):

skinkie:
Can we please have an example for this too? Preferably without the deprecated one.

OJP/OJP_FareSupport.xsd
```diff
@@ -446,11 +460,18 @@
...
 			</xs:element>
-			<xs:element name="BookingNote" type="InternationalTextStructure" minOccurs="0">
-				<xs:annotation>
-					<xs:documentation>Note about booking the FLEXIBLE LINE.</xs:documentation>
-				</xs:annotation>
-			</xs:element>
+			<xs:choice>
+				<xs:element name="BookingNote" type="BookingNoteType" minOccurs="0">
+					<xs:annotation>
+						<xs:documentation>Note about booking the LINE (deprecated).</xs:documentation>
```
sgrossberndt, line 466: If we are doing all kind of breaking the API that hard, why this deprecation here?
The BookingNote was introduced with 4e90159b "Aligning BookingArrangements more closely with NeTEx", which is not part of a published version. I opt to remove the choice and BookingNote completely in favor of only a BookingNotes element

sgrossberndt (changes requested):

OJP/OJP_FareSupport.xsd
```diff
@@ -8,6 +8,20 @@
...
+		<xs:annotation>
+			<xs:documentation>A structure for booking notes.</xs:documentation>
+		</xs:annotation>
+		<xs:sequence>
+			<xs:element name="BookingNote" type="BookingNoteType" minOccurs="1" maxOccurs="unbounded"/>
+		</xs:sequence>
+	</xs:complexType>
+	<xs:complexType name="BookingNoteType">
+		<xs:annotation>
+			<xs:documentation>An element of the BookingNotes or as a single BookingNote.</xs:documentation>
```
sgrossberndt, line 21: I do not understand this documentation. Could it please be reworded?

OJP/OJP_FareSupport.xsd
```diff
@@ -8,6 +8,20 @@
...
+		</xs:annotation>
+		<xs:attribute name="order" type="xs:positiveInteger"/>
```
sgrossberndt, line 23: Please do not add `xs:attribute`s.
Is a `order` really necessary and thus having the client re-order the notes? The response should contain the notes in correct order - for trips and other places we are able to do so as well..

sgrossberndt (changes requested):

OJP/OJP_FareSupport.xsd
ue71603: @skinkie no order for you ok then?

skinkie: @sgrossberndt lets discuss this afternoon. I don't like the implicit ordering idea.

OJP/OJP_FareSupport.xsd
ue71603: reworded

OJP/OJP_FareSupport.xsd
ue71603: exists in InternationalTextStructure already.

OJP/OJP_FareSupport.xsd
ue71603: adapted

OJP/OJP_FareSupport.xsd
```diff
@@ -8,6 +8,14 @@
 	<xs:annotation>
 		<xs:documentation>========================================== Fare Authority ==========================================</xs:documentation>
 	</xs:annotation>
+	<xs:complexType name="BookingNotesStructure">
+		<xs:annotation>
+			<xs:documentation>A structure for booking notes.</xs:documentation>
+		</xs:annotation>
+		<xs:sequence>
+			<xs:element name="BookingNote" type="InternationalTextStructure" minOccurs="1" maxOccurs="unbounded"/>
```
ue71603, line 16:
```suggestion
			<xs:element name="BookingNote" type="InternationalTextStructure" minOccurs="1" maxOccurs="unbounded">
			<xs:annotation>
			  <xs:documentation>The BookingNote should be treated as ordered.</xs:documentation>
			</xs:annotation>
			</xs:element>
```

sgrossberndt (approved):

OJP/OJP_FareSupport.xsd
sgrossberndt: order was removed

skinkie:
One question remaining: @Aurige mentioned these could be Notices with a NoticeAssignment. Is there any BookingNote in SIRI, NeTEx, that would justify it to create a new element?

ue71603:
NeTEx has Notice and NoticeAssignment
SIRI: does not have an Assignment thing. It has many simple Note types (NaturalLanguageStringStructure)
   * StopNotice & StopNoticeCancellation
   * StopLineNotice StopLineNoticeCancellation
   * Notice 
   * ChangeNote
   * StopVisitNote
   * LineNote
   * StopNote
   * VehicleActivityNote
   * CallNote
   * CallNoteGroup
   * StopVisitNote
   * JourneyNote

ue71603:
This would need to be fixed in SIRI first. The bug I already put in a PR https://github.com/SIRI-CEN/SIRI/pull/116

trurlurl (approved):

skinkie (approved):

PR #349, opened 2023-04-07
Enhance TripPolicy for more Private Lines, filter for PassengerCategory, Access Modes

ue71603:
With this PR it is possible to allow a trip planner to include specialised private LINE (e.g. car pooling, school busses, company shuttles) into the trip planning. 
It also allows to define necessary ACCESS MODE (e.g. that you are in a car, which restricts search to road, ferry, carTransportRail or that you can access a dragLift, because you have skis with you.
The third part makes sure, that selected by the PassengerCategory some additional things might get "unlocked". E.g. in Delemont seniors have a different service offered to them by the demand responsive service: https://www.postauto.ch/de/fahrplan-und-netz/publicar/delemont

PR #379, opened 2023-07-03
implementing MODES OF OPERATION from FU 44

ue71603:
### 
image

image

Additionally, The association from NAVIGATION PATH to ACCESS MODE should be removed and replaced with an association between NAVIGATION PATH and PERSONAL MODE.

Maybe:
NAVIGATION PATH --> traversable by --> PERSONAL MODE
PERSONAL MODE  --> suitable for      --> NAVIGATION PATH 
image

### Current state of affairs
MODE OJP

List in Transmodel
dynamicCarPooling
longDistanceCarPooling
stationlessVehicleSharing
peerToPeerCarSharing
taxicab
sharedTaxi
commercialCarSharing
carSharingClub
cycleSharing
chauffeuredCar
taxi

netex_modeOfOperation_support

Modes of Operation:
scheduled
flexible
alternative
vehicleRental
vehicleSharing
vehiclePooling
personal
other

Enums of SubModesOfOperation:

flexible
			<xsd:enumeration value="flexibleRoute"/>
			<xsd:enumeration value="flexibleArea"/>
			<xsd:enumeration value="demandResponsive"/>
rental:
			<xsd:enumeration value="vehicleHire"/>
			<xsd:enumeration value="vehicleLease"/>
sharing:
			<xsd:enumeration value="carSharingClub"/>
			<xsd:enumeration value="peerToPeerCarSharing"/>
			<xsd:enumeration value="vehicleSharing"/>
pooling
			<xsd:enumeration value="taxi"/>
			<xsd:enumeration value="appBookedTaxi"/>
			<xsd:enumeration value="sharedTaxi"/>
			<xsd:enumeration value="prebookedTaxi"/>
			<xsd:enumeration value="chaffeuredVehicle"/>
			<xsd:enumeration value="dynamicCarPooling"/>
			<xsd:enumeration value="longDistanceCarPooling"/>
personal
			<xsd:enumeration value="commuterCarPooling"/>
			<xsd:enumeration value="ownCar"/>
			<xsd:enumeration value="privateLift"/>
scheduled
			<xsd:enumeration value="scheduledService"/>
			<xsd:enumeration value="eventBasedService"/>


OJP/OJP_ModesSupport.xsd
```diff
@@ -8,19 +8,43 @@
...
 		</xs:annotation>
 		<xs:restriction base="xs:string">
 			<xs:enumeration value="walk"/>
 			<xs:enumeration value="cycle"/>
-			<xs:enumeration value="taxi"/>
 			<xs:enumeration value="self-drive-car"/>
 			<xs:enumeration value="others-drive-car"/>
+			<xs:enumeration value="vehicleHire">
+				<xs:annotation>
+					<xs:documentation>One of the MODES OF OPERATIOn from NeTEX rental.</xs:documentation>
```
skinkie, line 22:
```suggestion
					<xs:documentation>One of the MODES OF OPERATION from NeTEX rental.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -8,19 +8,43 @@
...
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="vehicleLease">
+				<xs:annotation>
+					<xs:documentation>One of the MODES OF OPERATIOn from NeTEX rental.</xs:documentation>
```
skinkie, line 27:
```suggestion
					<xs:documentation>One of the MODES OF OPERATION from NeTEX rental.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -8,19 +8,43 @@
...
+				</xs:annotation>			
+			</xs:enumeration>			
 			<xs:enumeration value="motorcycle"/>
 			<xs:enumeration value="truck"/>
 			<xs:enumeration value="scooter"/>
+			<xs:enumeration value="carSharingClub">
+				<xs:annotation>
+					<xs:documentation>Car sharing club. MODES OF OPERATIOn from NeTEx sharing.</xs:documentation>
```
skinkie, line 35:
```suggestion
					<xs:documentation>Car sharing club. MODES OF OPERATION from NeTEx sharing.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -8,19 +8,43 @@
...
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="peerToPeerCarSharing">
+				<xs:annotation>
+					<xs:documentation>Sharing between peers. MODES OF OPERATIOn from NeTEx sharing.</xs:documentation>
```
skinkie, line 40:
```suggestion
					<xs:documentation>Sharing between peers. MODES OF OPERATION from NeTEx sharing.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -8,19 +8,43 @@
...
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="vehicleSharing">
+				<xs:annotation>
+					<xs:documentation>Non-car (bicycles, scooters and the like) sharing. MODES OF OPERATIOn from NeTEx sharing.</xs:documentation>
```
skinkie, line 45:
```suggestion
					<xs:documentation>Non-car (bicycles, scooters and the like) sharing. MODES OF OPERATION from NeTEx sharing.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -78,16 +108,41 @@
 			</xs:element>
 		</xs:sequence>
 	</xs:complexType>
-	<xs:simpleType name="ContinuousModesEnumeration">
+	<xs:simpleType name="ContinuousModesOfOperationEnumeration">
 		<xs:annotation>
-			<xs:documentation>Types of cmodes that  run at any time without a timetable.</xs:documentation>
+			<xs:documentation>Types of MODES OF OPERATION  that  run at any time without a timetable and that are not classic flexible services.</xs:documentation>
```
skinkie, line 113:
```suggestion
			<xs:documentation>Types of MODES OF OPERATION that run at any time without a timetable and that are not classic flexible services.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
ue71603: I even copy&paste my errors.

OJP/OJP_ModesSupport.xsd
```diff
@@ -78,16 +108,41 @@
...
-	<xs:simpleType name="ContinuousModesEnumeration">
+	<xs:simpleType name="ContinuousModesOfOperationEnumeration">
 		<xs:annotation>
-			<xs:documentation>Types of cmodes that  run at any time without a timetable.</xs:documentation>
+			<xs:documentation>Types of MODES OF OPERATION that run at any time without a timetable and that are not classic flexible services.</xs:documentation>
 		</xs:annotation>
 		<xs:restriction base="xs:string">
 			<xs:enumeration value="walk"/>
-			<xs:enumeration value="demandResponsive"/>
-			<xs:enumeration value="replacementService"/>
```
skinkie, line 88: Where are those two going?

ue71603: demandResponsive is now in TimedMode. replacementService is gone. Was not on the list. I check.

OJP/OJP_ModesSupport.xsd
skinkie: At least your MODE OF OPERATION is consistent ;)

OJP/OJP_ModesSupport.xsd
ue71603: replacementService now also in TimedMode....

ue71603:
Things we don't want to do are now in: https://github.com/VDVde/OJP/issues/380

OJP/OJP_ModesSupport.xsd
herlitze: If I remember correctly, we will have another discussion about demandResponsive being part of ContinuousModesOfOperationEnumeration,  or adding the GeoPosition to TimedLegs. I am in favor of adding it back to ContinuousModesOfOperationEnumeration.

herlitze (commented):
We have to wait for our discussion and the solution for demandResponsive

OJP/OJP_ModesSupport.xsd
ue71603: @herlitze  We have a problem there: Then area based demand responsive is a ContinuousLeg and not a TimedLeg. And I think it was also you who suggested that we should use a TimedLeg for every type of demand responsive public transport.  Can you clarify?

herlitze: @ue71603 I remember it that way that I agreed, but I was not the one who suggested removing demand responsive from ContinousLeg. We were not aware of the problems of mapping demand responsive transport serving addresses in a TimedLeg. With this UseCase we have to rethink the decision.

In the current state of the PullRequest we can pass a GeoPosition in addition to the existing StopPointGroup. If I recall our discussion correctly, if we want to keep demand responsive only in TimedLegs, we will need to support addresses as well, so the whole Place-structure is needed.
I dislike the idea, but the plan was for us to talk about the problem again, right?

OJP/OJP_ModesSupport.xsd
```diff
@@ -8,19 +8,43 @@
 	<xs:annotation>
 		<xs:documentation>========================================== Individual Modes ==========================================</xs:documentation>
 	</xs:annotation>
-	<xs:simpleType name="IndividualModesEnumeration">
+	<xs:simpleType name="IndividualModesOfOperationEnumeration">
 		<xs:annotation>
 			<xs:documentation>modes which an individual powers themselves (such as walk, cycle)</xs:documentation>
```
Aurige, line 13: "Powers" or "Drives" ?

OJP/OJP_ModesSupport.xsd
```diff
@@ -30,18 +54,24 @@
 	</xs:simpleType>
 	<xs:simpleType name="PrivateModesEnumeration">
 		<xs:annotation>
-			<xs:documentation>[a category of MODE in TMv6] MODEs offered by private individuals </xs:documentation>
+			<xs:documentation>MODE OF OPERATION offered by private individuals (contains the values as defined by NeTEx). In the end these are all different forms of car pooling</xs:documentation>
```
Aurige, line 57: so why not PrivateModesOf OperationEnumeration (as we have IndividualModesOfOperationEnumeration)

OJP/OJP_ModesSupport.xsd
```diff
@@ -78,12 +108,38 @@
...
 	</xs:complexType>
-	<xs:simpleType name="ContinuousModesEnumeration">
+	<xs:simpleType name="ContinuousModesOfOperationEnumeration">
 		<xs:annotation>
-			<xs:documentation>Types of cmodes that  run at any time without a timetable.</xs:documentation>
+			<xs:documentation>Types of MODES OF OPERATION that run at any time without a timetable and that are not classic flexible services.</xs:documentation>
 		</xs:annotation>
 		<xs:restriction base="xs:string">
 			<xs:enumeration value="walk"/>
+			<xs:enumeration value="commuterCarPooling">
```
Aurige, line 117: Why is this Continuous: there is most often a scheduled pick up time !

Aurige (commented):
Sorry I'm late to this discussion, and try to catch-up

To Malte: demandResponsive is not always Continuous; for example Virtual Lines are demandResponsive and have schedules. You can't really link them as both situation occur. And to me that sounds really weird to mess the category just because the attribute we need (GeoPosition) is only in one of these categories.

One other thing I find a bit confusing is that IndividualModesOfOperationEnumeration and PrivateModesEnumeration ere in fact compbinations of (vehicle)mode and mode of operation. Also in PrivateModesEnumeration we have things like taxi/appBookedTaxi/sharedTaxi/prebookedTaxi but a single taxi service (mode of operation) can offer all these services (plus hailed taxi, but I guess that this is the "taxi" value), so we are mixing the mode of operation and the booking&service description.

I know that the NeTEx MODEs are far from perfect and often blurry, but I have the feeling that we are replacing a blurry sition by another blurry one.

OJP/OJP_ModesSupport.xsd
ue71603, line 13:
```suggestion
			<xs:documentation>modes which an individual "drives"  themselves and walk (such as walk, cycle)</xs:documentation>
```

ue71603: changed

OJP/OJP_ModesSupport.xsd
ue71603: @Aurige  comes with the pickup being a PLACE and not a SITE ELEMENT. We don't have the tools for that in OJP. So perhaps you want to weight in on that.  Perhaps we need to have some enum values in both enumerations "demandresponsive" for TimedLeg (line based) and ContinuousLeg (for area non-stop based). This might also be true for commutercarPooling.  Should I update accordingly?

OJP/OJP_ModesSupport.xsd
ue71603: changed in the next commit.

ue71603:
@trurlurl @herlitze @normanoffel @Aurige : 
What we need to do next week:
* Check that we added the things from Transmodel FU44 to the definitions in OJP.  Christophe/Matthias
* Make clear that we will only have SITE ELEMENTS in TimedLegs => area based door-2-door demand responsive buses are ContinuousLegs
* Cleanup the enums (I think they are currently not clear cut)

ue71603:
[MODE_OF_OPERATION.xlsx](https://github.com/VDVde/OJP/files/12659534/MODE_OF_OPERATION.xlsx)
@skinkie  @herlitze @normanoffel  @Aurige   @sgrossberndt  @trurlurl 
I tried to map MODE OF OPERATION, MODE and all the rest into one consistent picture:
* We keep TimedLeg, TransferLeg, ContinuousLeg
* We have a ModeOfOperation (which bases on xxxService currently). There are many and they may affect, which MODE / SUBMODE can then be used. But we don't explicitly model that.
* I suggest we remove the current different types of xxxModeOfOperation for TimedLeg (Service covers this just fine)
* As TransferLeg has no Service I will have to tweak TransferMode (there are some strange modes in there), ContinousMode will be integrated in the TransferMode. I think TransferModeOfOperation is not needed explicitly.
* and ContinuousLeg, if possible and only keep it for the TransferlegI suggest we remove the current different types of xxxModeOfOperation for TimedLeg and ContinuousLeg, if possible and only keep it for the Transferleg
* In the ConinuousServiceStructure, I will have to twak things: ContinousMode/IndividualMode will be merged. ModeOfOperation needs to be added and how to handle the DatedVehicleJourney vs SharingService will need to be checked

Can you pls check quickly and give me your input on that part?

trurlurl:
> Can you pls check quickly and give me your input on that part?

The plan sounds good to me, as far as I understand it: 

- TimedLeg would have no xxxModes. From TimedModes, flexibleArea would move to ContinuousModes, I think.
- TransferLeg would have TransferModes, perhaps complemented with items from ContinuousModes (would car or motorcycle make sense, or would such a motorised transfer be modelled as ContinuousLeg?)
- ContinuousLeg would have ContinuousModes that includes flexibleArea and items from IndividualModes. And an additional ContinuousModeOfOperation or similar.

ue71603:
> > Can you pls check quickly and give me your input on that part?
> 
> The plan sounds good to me, as far as I understand it:
> 
> * TimedLeg would have no xxxModes. From TimedModes, flexibleArea would move to ContinuousModes, I think.
> * TransferLeg would have TransferModes, perhaps complemented with items from ContinuousModes (would car or motorcycle make sense, or would such a motorised transfer be modelled as ContinuousLeg?)
> * ContinuousLeg would have ContinuousModes that includes flexibleArea and items from IndividualModes. And an additional ContinuousModeOfOperation or similar.

Yes, that's about it.

OJP/OJP_ModesSupport.xsd
```diff
@@ -8,31 +8,62 @@
 	<xs:annotation>
 		<xs:documentation>========================================== Individual Modes ==========================================</xs:documentation>
 	</xs:annotation>
-	<xs:simpleType name="IndividualModesEnumeration">
+	<xs:simpleType name="IndividualModesOfOperationEnumeration">
 		<xs:annotation>
-			<xs:documentation>modes which an individual powers themselves (such as walk, cycle)</xs:documentation>
+			<xs:documentation>modes which an individual "drives"  themselves and walk (such as walk, cycle)</xs:documentation>
```
skinkie, line 13:
```suggestion
			<xs:documentation>modes which an individual "drives" themselves and walk (such as walk, cycle)</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -82,12 +113,43 @@
...
 		<xs:restriction base="xs:string">
 			<xs:enumeration value="walk"/>
+			<xs:enumeration value="commuterCarPooling">
+				<xs:annotation>
+					<xs:documentation>MODE OF OPERATION from NeTEX personal.</xs:documentation>
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="demandResponsive">
+				<xs:annotation>
+					<xs:documentation>area based demand responsive service (without SCHEDULED STOP POINTs.</xs:documentation>
```
skinkie, line 129:
```suggestion
					<xs:documentation>area based demand responsive service (without SCHEDULED STOP POINTs).</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -82,12 +113,43 @@
...
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="ownCar">
+				<xs:annotation>
+					<xs:documentation>MODE OF OPERATION from NeTEX personal.</xs:documentation>
```
skinkie, line 134:
```suggestion
					<xs:documentation>MODE OF OPERATION from NeTEx personal.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -82,12 +113,43 @@
...
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="privateLift">
+				<xs:annotation>
+					<xs:documentation>MODE OF OPERATION from NeTEX personal.</xs:documentation>
```
skinkie, line 139:
```suggestion
					<xs:documentation>MODE OF OPERATION from NeTEx personal.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -8,31 +8,62 @@
...
 			<xs:enumeration value="self-drive-car"/>
 			<xs:enumeration value="others-drive-car"/>
+			<xs:enumeration value="vehicleHire">
+				<xs:annotation>
+					<xs:documentation>One of the MODES OF OPERATION from NeTEX rental.</xs:documentation>
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="vehicleLease">
+				<xs:annotation>
+					<xs:documentation>One of the MODES OF OPERATION from NeTEX rental.</xs:documentation>
```
skinkie, line 27:
```suggestion
					<xs:documentation>One of the MODES OF OPERATION from NeTEx rental.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -82,12 +113,43 @@
...
+	<xs:simpleType name="ContinuousModesOfOperationEnumeration">
 		<xs:annotation>
-			<xs:documentation>Types of cmodes that  run at any time without a timetable.</xs:documentation>
+			<xs:documentation>Types of MODES OF OPERATION that run at any time without a timetable and that are not classic flexible services.</xs:documentation>
 		</xs:annotation>
 		<xs:restriction base="xs:string">
 			<xs:enumeration value="walk"/>
+			<xs:enumeration value="commuterCarPooling">
+				<xs:annotation>
+					<xs:documentation>MODE OF OPERATION from NeTEX personal.</xs:documentation>
```
skinkie, line 124:
```suggestion
					<xs:documentation>MODE OF OPERATION from NeTEx personal.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -8,31 +8,62 @@
...
 		</xs:annotation>
 		<xs:restriction base="xs:string">
 			<xs:enumeration value="walk"/>
 			<xs:enumeration value="cycle"/>
-			<xs:enumeration value="taxi"/>
 			<xs:enumeration value="self-drive-car"/>
 			<xs:enumeration value="others-drive-car"/>
+			<xs:enumeration value="vehicleHire">
+				<xs:annotation>
+					<xs:documentation>One of the MODES OF OPERATION from NeTEX rental.</xs:documentation>
```
skinkie, line 22:
```suggestion
					<xs:documentation>One of the MODES OF OPERATION from NeTEx rental.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -82,12 +113,43 @@
...
+			<xs:enumeration value="privateLift">
+				<xs:annotation>
+					<xs:documentation>MODE OF OPERATION from NeTEx personal.</xs:documentation>
+				</xs:annotation>
+			</xs:enumeration>
+		</xs:restriction>
+	</xs:simpleType>
+	<xs:simpleType name="TimedModesOfOperationEnumeration">
+		<xs:annotation>
+			<xs:documentation>Types of MODES OF OPERATION  are scheduled, classic and also flexible. From NeTEx.</xs:documentation>
```
skinkie, line 146: Do you want to mention where from NeTEx this is here too?

ue71603:
@skinkie  @Aurige  ready to have a look at it from the TM perspective 

OJP will implement the following MODE OF OPERATION. The hierarchy is not explicitly modeled. However, always the most specialised MODE OF OPERATION is to be used. 

Conventional  (ConventionalModesOfOperationEnumeration in TimedLeg)
- scheduled
- demandResponsive
  - flexibleRoute
  - flexibleArea
- shuttle
- pooling (when a timetable exists and SCHEDULED STOP POINTs exist)
- replacement
- pRM  (shuttles that are used instead of regular buses, when special needs warrant it)

Personal  (PersonalModesOfOperationEnumeration in ContinuousLeg)
- self  (= walking)
- own
- lease
- hire
- other

Alternative  (AlternativeModesOfOperation in ContinuousLeg)
- sharing
  - stationLessSharing
  - stationBasedSharing
  - peerToPeerSharing
  - parkAndRideSharing
  - sharingClubSharing
- pRM (when the operation is not based on SCHEDULED STOP POINTS)
pooling
  - taxi
  - shuttle
  - dynamicPooling
  - longDistancePooling
  - commuterPooling
  - parkAndRidePooling
  - chauffeured
- demandResponsive (flexible areas without stops)
  - flexibleArea
- other


OJP currently uses TransferModesEnumeration. However, given the values and the reason that MODE and TYPE are convoluted. We will use TransferType and 

Values:
- walk
- shuttle
- taxi
- protectedConnection
- guaranteedConnection
- remainInVehicle
- changeWithinVehicle
- checkIn
- checkOut
- parkAndRide
- bikeAndRide
- carHire
- bikeHire
- other

and I had to create an 

IndividalModeEnumeration
image

ue71603:
[MODE_OF_OPERATION.xlsx](https://github.com/VDVde/OJP/files/12689487/MODE_OF_OPERATION.xlsx)

OJP/OJP_Trips.xsd
```diff
@@ -120,9 +120,9 @@
 					</xs:annotation>
 				</xs:element>
 				<xs:sequence>
-					<xs:element name="ItModesToCover" type="IndividualModesEnumeration" minOccurs="0" maxOccurs="unbounded">
+					<xs:element name="ItModesToCover" type="PersonalModesOfOperationEnumeration" minOccurs="0" maxOccurs="unbounded">
 						<xs:annotation>
-							<xs:documentation>For each mode in this list a separate monomodal trip shall be found - in addition to inter-modal solutions.</xs:documentation>
+							<xs:documentation>For each mode in this list a separate monomodal trip shall be found - in addition to inter-modal solutions.XXXTODO </xs:documentation>
```
ue71603, line 125:
```suggestion
							<xs:documentation>For each mode in this list a separate monomodal trip shall be found - in addition to inter-modal solutions.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
ue71603: changed

OJP/OJP_ModesSupport.xsd
ue71603: updated

OJP/OJP_ModesSupport.xsd
ue71603: no longer referencing NeTEx but the Transmodel discussion.

Aurige:
> @skinkie @Aurige ready to have a look at it from the TM perspective

Looks ok for me (I didn't went through a detailed review of all XSD changes, but it looks fine).

OJP/OJP_Trips.xsd
```diff
@@ -120,7 +120,7 @@
 					</xs:annotation>
 				</xs:element>
 				<xs:sequence>
-					<xs:element name="ItModesToCover" type="IndividualModesEnumeration" minOccurs="0" maxOccurs="unbounded">
+					<xs:element name="ItModesToCover" type="PersonalModesOfOperationEnumeration" minOccurs="0" maxOccurs="unbounded">
```
trurlurl, line 123: Shouldn' t that now become `IndividualModeEnumeration` (without "s") rather than `PersonalModesOfOperationEnumeration` - which doesn't contain things like car, cycle, or taxi any more?

examples/functions/trip/TR_06_response_private_sharing.xml
```diff
@@ -73,9 +73,18 @@
 									<Order>15</Order>
 								</LegAlight>
 								<Service>
+									<OperatingDayRef>2023-08-13</OperatingDayRef>
 									<JourneyRef>1231231231:1231:123123</JourneyRef>
-									<PrivateMode>schoolBus</PrivateMode>
```
trurlurl, line 77: I think `PrivateMode` doesn't exist any more? Though my feeling is that we should keep it.

OJP/OJP_Common.xsd
```diff
@@ -283,11 +283,12 @@
 			</xs:element>
 		</xs:sequence>
 	</xs:group>
-	<xs:complexType name="SharingServiceStructure">
+	<xs:complexType name="AlternativeServiceStructure">
 		<xs:annotation>
 			<xs:documentation>Service that provides shared vehicles (The Transmodel MODE OF OPERATION is VEHICLE SHARING; other related Transmodel concepts: ALTERNATIVE MODE LEG.SINGLE JOURNEY.COMMON.VEHICLE SERVICE.TRANSPORT ORGANISATION).</xs:documentation>
 		</xs:annotation>
 		<xs:sequence>
+			<xs:element name="Mode" type="IndividualModeEnumeration"/>
```
trurlurl, line 291: This doesn't allow for "schoolBus", "companyShuttle", etc. Below there is the `Restricted` element, so such private modes should be handled somewhere.

OJP/OJP_Common.xsd
```diff
@@ -283,11 +283,12 @@
 			</xs:element>
 		</xs:sequence>
 	</xs:group>
-	<xs:complexType name="SharingServiceStructure">
+	<xs:complexType name="AlternativeServiceStructure">
```
trurlurl, line 286: See below the `SharingModel` element and structure - do we still need them, given "stationLess" etc. in `AlternativeModesOfOperationEnumeration`?

ue71603: Currently it is a bit more detailed. So I would leave it in the system. If the group would like to have it removed, I can certainly do that. @skinkie  @Aurige  @herlitze

OJP/OJP_Common.xsd
ue71603: This works, because 
ContinuousServiceStructure has a choice between
* AlternativeModeOfOperation or
* PersonalModeOfOperation

before there is the choice between:
* DatedJourneyGroup
* IndividualMode
* AlternativeService

In combination between
AlternativeService/Mode   and AlternativeModeOfOperation I think all these should be expressable. Or do you think something is missing?  @trurlurl

OJP/OJP_Trips.xsd
ue71603: We have the s in all Enums. I would keep it, or is there opposition? @Aurige @skinkie  @trurlurl

examples/functions/trip/TR_06_response_private_sharing.xml
ue71603: will check this

OJP/OJP_Trips.xsd
skinkie: You can keep the 's', consistency is key.

OJP/OJP_Common.xsd
skinkie: I am open to keeping and/or removing. No show stopper.

OJP/OJP_Common.xsd
trurlurl: I see that the use of a "companyShuttle" and other "private modes" can be expressed as a `AlternativeModeOfOperation" elsewhere. But then, in my view, one should remove all mentions of "private modes" and "PrivateMode" (e.g., in `PrivateServiceGroup` or in the annotation of `IncludeAllRestrictedLines`, see #349). Private modes wouldn't exist any more, only "private modes of operation". And perhaps that wouldn't be needed any more either - just speaking of of modes of operation that may be restricted somehow, as will be signalled in a `RestrictionNote`?

As long as one uses the term "private mode", I personally would expect to find it here in the element `Mode`. That was my problem.

OJP/OJP_Trips.xsd
trurlurl: I agree that having the "s" would be the correct solution. But that's already a step further than what I was stumbling across. What I wanted to say: I think `ItModesToCover` should not be of type `PersonalModesOfOperationEnumeration`, but of type `IndividualModeOperation` as defined elsewhere in this PR.

examples/functions/trip/TR_06_response_private_sharing.xml
trurlurl: My feeling came from the fact that "private modes" is spoken of at various places and in the context of the recent #349, see my other comment above. I don't see any other problem.

OJP/OJP_Common.xsd
herlitze: For me it is also fine to leave the structure.

OJP/OJP_Trips.xsd
```diff
@@ -674,12 +674,12 @@
 		</xs:annotation>
 		<xs:sequence>
 			<xs:choice>
-				<xs:element name="TransferMode" type="TransferModesEnumeration">
+				<xs:element name="TransferType" type="TransferTypeEnumeration">
```
herlitze, line 677: TransferTypeEnumeration includes parkAndRide and BikeAndRide, how does this work?

OJP/OJP_Trips.xsd
```diff
@@ -674,12 +674,12 @@
...
 					<xs:annotation>
-						<xs:documentation>Mode that is used for this interchange between public services.</xs:documentation>
+						<xs:documentation>TYPE that is used for this interchange between public services (TYPE OF TRANSFER).</xs:documentation>
 					</xs:annotation>
 				</xs:element>
-				<xs:element name="ContinuousMode" type="ContinuousModesEnumeration">
+				<xs:element name="AlternativeModeOfOperation" type="AlternativeModesOfOperationEnumeration">
```
herlitze, line 682: I don't see how parkAndRideSharing, longDistancePooling, commuterPooling, parkAndRidePooling, hotelShuttle, companyShuttle, school work for a tranferLeg.

herlitze:
> Updated the Excel. One column was misleading. [MODE_OF_OPERATION.xlsx](https://github.com/VDVde/OJP/files/12697795/MODE_OF_OPERATION.xlsx)

The table shows the mapping of Transmodel to OJP, correct?
I think a table mapping the current OJP modes to the new ones would be helpful. 
E. g. we added ride-pool-car, car-sharing, cycle-sharing, scooter-sharing, park-ride-car to the individualModes. These are now modelled as ContinuosLeg with Service.IndividualMode=car/cycle/scooter and Service.AlternativeModeOfOperation=sharing/pooling

trurlurl:
Where is the former distinction between self-drive car, others-drive, park-and-ride made - at least as far as it is necessary to know whether a parking space is needed or not?

OJP/OJP_Trips.xsd
ue71603: I think that having to park something takes additional time. That was the idea. You disagree?

OJP/OJP_Trips.xsd
ue71603: OK. will work on the lists again.

OJP/OJP_Common.xsd
ue71603: will go through it.

OJP/OJP_Trips.xsd
herlitze: I don't disagree with the additional time. P&R and B&R in my understanding is that I drive my own car/bike to a parking lot connected to a stop, leave it there, and then transfer to public transit. So it would only make sense as a first mile. How does it work as a TransferLeg?

OJP/OJP_Availability.xsd
```diff
@@ -143,7 +143,7 @@
 			</xs:element>
 			<xs:element name="WalkingImpaired" type="xs:boolean" minOccurs="0">
 				<xs:annotation>
-					<xs:documentation>Passenger is (maybe temporarily) walking impaired. Default is FALSE.</xs:documentation>
+					<xs:documentation>Passenger is (maybe temporarily) footing impaired. Default is FALSE.</xs:documentation>
```
ue71603, line 146:
```suggestion
					<xs:documentation>Passenger is (maybe temporarily) walking impaired. Default is FALSE.</xs:documentation>
```

OJP/OJP_JourneySupport.xsd
```diff
@@ -504,20 +509,21 @@
...
 					<xs:annotation>
-						<xs:documentation>Continuous transport options.</xs:documentation>
+						<xs:documentation>Alternative transport options.</xs:documentation>
 					</xs:annotation>
 				</xs:element>
-				<xs:element name="IndividualMode" type="IndividualModesEnumeration">
+				<xs:element name="PersonalModeOfOperation" type="PersonalModesOfOperationEnumeration" minOccurs="0" default="self">
 					<xs:annotation>
-						<xs:documentation>Individual transport options.</xs:documentation>
+						<xs:documentation>Personaltransport options.</xs:documentation>
```
ue71603, line 519:
```suggestion
						<xs:documentation>Personal transport options.</xs:documentation>
```

OJP/OJP_Trips.xsd
ue71603: @herlitze  wouldn't you  model  ?? as a TransferLeg?   car -> car parking at station --> ?? --> station    . Especially if it is a park garage or something like that. Pls advise. and then I update the model

OJP/OJP_Common.xsd
```diff
@@ -298,15 +289,20 @@
 					<xs:documentation>Public name of the service.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
-			<xs:element name="SharingModel" type="SharingModelEnumeration" minOccurs="0">
+			<xs:group ref="AlternativeServiceUsageGroup" minOccurs="0"/>
+			<xs:element name="InfoUrl" type="WebLinkStructure" minOccurs="0">
 				<xs:annotation>
-					<xs:documentation>Type of the sharing scheme.</xs:documentation>
+					<xs:documentation>Link to web page providing more details on service.</xs:documentation>
```
skinkie, line 295:
```suggestion
					<xs:documentation>Link to the web page providing more details on the service.</xs:documentation>
```

OJP/OJP_JourneySupport.xsd
```diff
@@ -224,13 +229,13 @@
 			<xs:group ref="ScheduledJourneyGroup"/>
 		</xs:sequence>
 	</xs:group>
-	<xs:group name="PrivateServiceGroup">
+	<!--	<xs:group name="PrivateServiceGroup">
```
skinkie, line 232: If not in use, please remove it.

ue71603:
> Where is the former distinction between self-drive car, others-drive, park-and-ride made - at least as far as it is necessary to know whether a parking space is needed or not?

@trurlurl This is in  
* the xxxModeOfOperation for self-drive others-drive in Service  PersonalModeOfOperation / AlternativeModeOfOperation
* indicating park-and-ride  is in the TransferTypeEnumeration and so it is part of the TransferLeg
* We are currently lacking ParkingFacility in StopServiceStatusGroup/Attribute (part of https://github.com/VDVde/OJP/pull/406 ). II will add it there.

trurlurl:
> > Where is the former distinction between self-drive car, others-drive, park-and-ride made - at least as far as it is necessary to know whether a parking space is needed or not?
> 
> @trurlurl This is in
> 
>     * the xxxModeOfOperation for self-drive others-drive in Service  PersonalModeOfOperation / AlternativeModeOfOperation
> 
>     * indicating park-and-ride  is in the TransferTypeEnumeration and so it is part of the TransferLeg
> 
>     * We are currently lacking ParkingFacility in StopServiceStatusGroup/Attribute (part of [Added Attributes to LegBoard, LegAligt und LegIntermediate #406](https://github.com/VDVde/OJP/pull/406) ). II will add it there.

@ue71603 Do you mean by that that self-drive and others-drive are in PersonalModeOfOperation? There is "self" (=self-drive), but otherwise only "own", "other-owned", etc. (nothing indicating a pick-up or drop-off scenario). (BTW: are the ownership relations needed?)

And I don't understand why parkAndRide is on the TransferLeg. I would expect parkAndRide vs. pick-up/drop-off to be labelled on a ContinuousLeg (in PersonalModeOfOperation).

OJP/OJP_ModesSupport.xsd
```diff
@@ -1,55 +1,189 @@
...
+	<xs:annotation>
+		<xs:documentation>==========================================IndividualTransportOptionsStructure ==========================================</xs:documentation>
+	</xs:annotation>
 	<xs:complexType name="IndividualTransportOptionsStructure">
 		<xs:annotation>
 			<xs:documentation>Individual modes and their usage limits as stated by the passenger (TRIP ACCESS CONSTRAINT).</xs:documentation>
 		</xs:annotation>
 		<xs:sequence>
-			<xs:element name="Mode" type="IndividualModesEnumeration">
+			<xs:element name="PersonalModeOFOperation" type="PersonalModesOfOperationEnumeration" minOccurs="0">
```
trurlurl, line 177: The annotation will need to be updated, depending on the enum values we'll have in the end.

OJP/OJP_JourneySupport.xsd
```diff
@@ -211,46 +211,22 @@
...
-					<xs:documentation>Description of the private service.</xs:documentation>
-				</xs:annotation>
-			</xs:element>
-			<xs:element name="OrganisationRef" type="siri:OperatorRefStructure" minOccurs="0"/>
-			<xs:element name="InfoUrl" type="WebLinkStructure" minOccurs="0">
-				<xs:annotation>
-					<xs:documentation>Link to web page providing more details on service.</xs:documentation>
-				</xs:annotation>
-			</xs:element>
+			<xs:group ref="ScheduledJourneyGroup"/>
```
trurlurl, line 229: `ScheduledJouneyGroup` already contains an optional `siri:VehicleRef` - remove it above (currently line 238/224)?

ue71603:
@trurlurl 

> Do you mean by that that self-drive and others-drive are in PersonalModeOfOperation? There is "self" (=self-drive), but otherwise only "own", "other-owned", etc. (nothing indicating a pick-up or drop-off scenario). (BTW: are the ownership relations needed?)

No: other drive etc is always an AlternativeModeOfOperation. PersonalModeOfOperation is always yourself in your vehicle.

>And I don't understand why parkAndRide is on the TransferLeg. I would expect parkAndRide vs. pick-up/drop-off to be labelled on a ContinuousLeg (in PersonalModeOfOperation).

The park and ride happens in my opinion:
* ContinuousLeg you do with your car to the parking
* TransferLeg between the parking and the public transport  (or between the ContinuousLeg with the car and a different ContinuousLeg with "foot" as MODE.
* TimedLeg public transport.
The TransferLeg then accounts for the time that is needed to navigate the PARKING and get to the track where the train will depart.

Other opinions to model this?

herlitze:
@ue71603 to what extent must the change also be taken into account for the PtModeFilter in the TripRequest?

ue71603:
> @ue71603 I have the following remarks:
> 
> * To what extent must the change also be taken into account for the PtModeFilter in the TripRequest?
> * To what extent does this apply to the ITModeToCover? For example, how is a sharing ride requested?
> * TransferType is not clear to me yet, parkandRide/bikeandRide are for me e.g. no means of transport I can use when transferring

1. ptModeFilter. You are right. We need an update there as well. Is fixed in this commit here: https://github.com/VDVde/OJP/pull/379/commits/5d4251968a914bb364fbc6aa04f002f67f38c757
2. ITModeToCover: This includes all PersonalMode currently. So sharing could not be requested. I rebuild ItModeToCover with a Mode/ModeOfOperation combination, then it should work.  See commit: https://github.com/VDVde/OJP/pull/379/commits/820156513eb6caa988d38f3cbe5a566df17b313f (examples had to be fixed afterwards, too)
3. Even when TransferType was not fully added that way by me, I expand now the definitions. Pls check, if this makes sense @herlitze @Aurige @trurlurl  Done in this commit: https://github.com/VDVde/OJP/pull/379/commits/b0a669c1c55df3a57f6e148ae878a85db2b7f8df

ue71603:
André noted (and I think that Malte also goes in this direction). How do I specify, that I need park and ride.
This now can be done
`					<IndividualTransportOptions>
						<ItModeAndModeOfOperation>
							<PersonalMode>car</PersonalMode>
						</ItModeAndModeOfOperation>
						<MaxDistance>240</MaxDistance>
						<MaxDuration>PT20M</MaxDuration>
						<MinDistance>100</MinDistance>
						<MinDuration>PT1M</MinDuration>
						<Speed>150</Speed>
						<AdditionalTime>PT3M</AdditionalTime>
					</IndividualTransportOptions>
`
makes sure that the first leg is car and P&R should be considered. It is not forced as on street parking might be possible. Do you want an additional boolean to force P&R. We could do this by adding an additional value to either  personal mode or something else like that.

In the destination 
`						<ItModeAndModeOfOperation>
							<PersonalMode>car</PersonalMode>
							<AlternativeModeOfOperation>parkAndRideSharing</AlternativeModeOfOperation>
						</ItModeAndModeOfOperation>`
would do the trick.

But for what André wanted in the first example
`					<IndividualTransportOptions>
						<ItModeAndModeOfOperation>
							<PersonalMode>foot</PersonalMode>
						</ItModeAndModeOfOperation>
					</IndividualTransportOptions>
					<IndividualTransportOptions>
						<ItModeAndModeOfOperation>
							<PersonalMode>car</PersonalMode>
							<AlternativeModeOfOperation>taxi</AlternativeModeOfOperation>
						</ItModeAndModeOfOperation>
					</IndividualTransportOptions>	`

OJP/OJP_Trips.xsd
```diff
@@ -120,9 +120,9 @@
 					</xs:annotation>
 				</xs:element>
 				<xs:sequence>
-					<xs:element name="ItModeToCover" type="PersonalModesEnumeration" minOccurs="0" maxOccurs="unbounded">
+					<xs:element name="ItModeToCover" type="ItModesStructure" minOccurs="0" maxOccurs="unbounded">
 						<xs:annotation>
-							<xs:documentation>For each mode in this list a separate monomodal trip shall be found - in addition to inter-modal solutions.</xs:documentation>
+							<xs:documentation>For each mode / mode of operation combination in this list a separate monomodal trip shall be found - in addition to inter-modal solutions.</xs:documentation>
```
skinkie, line 125:
```suggestion
							<xs:documentation>For each MODE or MODE OF OPERATION combination in this list a separate monomodal trip shall be found - in addition to inter-modal solutions.</xs:documentation>
```

examples/functions/trip/TR_14_first_leg_car_and_no_parking_last_leg_walk_or_bike_known_stop.xml
```diff
@@ -0,0 +1,62 @@
...
+						</Name>
+					</PlaceRef>
+					<IndividualTransportOptions>
+						<ItModeAndModeOfOperation>
+							<PersonalMode>foot</PersonalMode>
+						</ItModeAndModeOfOperation>
+					</IndividualTransportOptions>
+					<IndividualTransportOptions>
+						<ItModeAndModeOfOperation>
+							<PersonalMode>cycle</PersonalMode>
```
trurlurl, line 51: I guess this would result in having 'bikeAndRide' on the TransferLeg - unless 'BikeTransport' were selected and a sophisticated trip planner would differentially treat such cases? 

The parking / no parking information would always belong to the TransferLeg and not be visible on the ContinuousLeg.

trurlurl (commented):
Looks like a good solution to me.

examples/functions/trip/TR_12_response_example_restricted_service.xml
```diff
@@ -0,0 +1,158 @@
...
+									<siri:StopPointRef>8581235</siri:StopPointRef>
+									<StopPointName>
+										<Text xml:lang="de">Delémont, rue du Poujet</Text>
+									</StopPointName>
+									<ServiceDeparture>
+										<TimetabledTime>2023-09-13T14:11:00Z</TimetabledTime>
+									</ServiceDeparture>
+									<Order>1</Order>
+								</LegBoard>
+								<LegIntermediates>
```
trurlurl, line 34:
```suggestion
								<LegIntermediate>
```

examples/functions/trip/TR_12_response_example_restricted_service.xml
```diff
@@ -0,0 +1,158 @@
...
+										<Text xml:lang="de">Delémont, La Promenade</Text>
+									</StopPointName>
+									<ServiceArrival>
+										<TimetabledTime>2023-09-13T14:12:00Z</TimetabledTime>
+									</ServiceArrival>
+									<ServiceDeparture>
+										<TimetabledTime>2023-09-13T14:12:00Z</TimetabledTime>
+									</ServiceDeparture>
+									<Order>2</Order>
+								</LegIntermediates>
```
trurlurl, line 46:
```suggestion
								</LegIntermediate>
```

examples/functions/trip/TR_12_response_example_restricted_service.xml
```diff
@@ -0,0 +1,158 @@
...
+								<LegIntermediates>
```
trurlurl, line 47:
```suggestion
								<LegIntermediate>
```

examples/functions/trip/TR_12_response_example_restricted_service.xml
```diff
@@ -0,0 +1,158 @@
...
+										<Text xml:lang="de">Delémont, Château-Musée</Text>
+									</StopPointName>
+									<ServiceArrival>
+										<TimetabledTime>2023-09-13T14:12:00Z</TimetabledTime>
+									</ServiceArrival>
+									<ServiceDeparture>
+										<TimetabledTime>2023-09-13T14:12:00Z</TimetabledTime>
+									</ServiceDeparture>
+									<Order>3</Order>
+								</LegIntermediates>
```
trurlurl, line 59:
```suggestion
								</LegIntermediate>
```

examples/functions/trip/TR_12_response_example_restricted_service.xml
```diff
@@ -0,0 +1,158 @@
...
+								<LegIntermediates>
```
trurlurl, line 60:
```suggestion
								<LegIntermediate>
```

examples/functions/trip/TR_12_response_example_restricted_service.xml
```diff
@@ -0,0 +1,158 @@
...
+										<Text xml:lang="de">Delémont, Hôtel de Ville</Text>
+									</StopPointName>
+									<ServiceArrival>
+										<TimetabledTime>2023-09-13T14:13:00Z</TimetabledTime>
+									</ServiceArrival>
+									<ServiceDeparture>
+										<TimetabledTime>2023-09-13T14:13:00Z</TimetabledTime>
+									</ServiceDeparture>
+									<Order>4</Order>
+								</LegIntermediates>
```
trurlurl, line 72:
```suggestion
								</LegIntermediate>
```

examples/functions/trip/TR_12_response_example_restricted_service.xml
```diff
@@ -0,0 +1,158 @@
...
+								<LegIntermediates>
```
trurlurl, line 73:
```suggestion
								<LegIntermediate>
```

examples/functions/trip/TR_12_response_example_restricted_service.xml
```diff
@@ -0,0 +1,158 @@
...
+										<Text xml:lang="de">Delémont, Théâtre</Text>
+									</StopPointName>
+									<ServiceArrival>
+										<TimetabledTime>2023-09-13T14:14:00Z</TimetabledTime>
+									</ServiceArrival>
+									<ServiceDeparture>
+										<TimetabledTime>2023-09-13T14:14:00Z</TimetabledTime>
+									</ServiceDeparture>
+									<Order>5</Order>
+								</LegIntermediates>
```
trurlurl, line 85:
```suggestion
								</LegIntermediate>
```

examples/functions/trip/TR_12_response_example_restricted_service.xml
```diff
@@ -0,0 +1,158 @@
...
+								<LegIntermediates>
```
trurlurl, line 86:
```suggestion
								<LegIntermediate>
```

examples/functions/trip/TR_12_response_example_restricted_service.xml
```diff
@@ -0,0 +1,158 @@
...
+										<Text xml:lang="de">Delémont, Le Ticle</Text>
+									</StopPointName>
+									<ServiceArrival>
+										<TimetabledTime>2023-09-13T14:15:00Z</TimetabledTime>
+									</ServiceArrival>
+									<ServiceDeparture>
+										<TimetabledTime>2023-09-13T14:15:00Z</TimetabledTime>
+									</ServiceDeparture>
+									<Order>6</Order>
+								</LegIntermediates>
```
trurlurl, line 98:
```suggestion
								</LegIntermediate>
```

examples/functions/trip/TR_12_response_example_restricted_service.xml
```diff
@@ -0,0 +1,158 @@
...
+								<LegIntermediates>
```
trurlurl, line 99:
```suggestion
								<LegIntermediate>
```

examples/functions/trip/TR_12_response_example_restricted_service.xml
```diff
@@ -0,0 +1,158 @@
...
+										<Text xml:lang="de">Delémont, Av. de la gare</Text>
+									</StopPointName>
+									<ServiceArrival>
+										<TimetabledTime>2023-09-13T14:16:00Z</TimetabledTime>
+									</ServiceArrival>
+									<ServiceDeparture>
+										<TimetabledTime>2023-09-13T14:16:00Z</TimetabledTime>
+									</ServiceDeparture>
+									<Order>7</Order>
+								</LegIntermediates>
```
trurlurl, line 111:
```suggestion
								</LegIntermediate>
```

trurlurl (changes requested):

OJP/OJP_ModesSupport.xsd
```diff
@@ -1,53 +1,182 @@
...
+				<xs:annotation>
+					<xs:documentation>Only use this value, when no other applies.</xs:documentation>
+				</xs:annotation>
+			</xs:enumeration>
 		</xs:restriction>
 	</xs:simpleType>
+	<xs:annotation>
+		<xs:documentation>==========================================IndividualTransportOptionsStructure ==========================================</xs:documentation>
+	</xs:annotation>
 	<xs:complexType name="IndividualTransportOptionsStructure">
```
trurlurl, line 172: Shouldn't it be rather a singular, the structure and also the elements: IndividualTransportOptionStructure, IndividualTransportOption? Since multiple IndividualTransportOption can be listed in sequence, each specifying one option/alternative.

trurlurl (changes requested):

OJP/OJP_ModesSupport.xsd
ue71603: done (even when it is not strictly part of this PR)

OJP/OJP_JourneySupport.xsd
```diff
@@ -599,11 +576,11 @@
 				</xs:element>
 				<xs:element name="TimeAllowance" type="xs:duration" minOccurs="0">
 					<xs:annotation>
-						<xs:documentation>Extra time needed before reaching/after leaving this location (an example of a TRIP ACCESS CONSTRAINT. In Transmodel it is modeled more extensively and could relate to ACCESS MODE e.g. walk max 5 mintues, but cycle 10 minutes).</xs:documentation>
+						<xs:documentation>Extra time needed before reaching/after leaving this location (an example of a TRIP ACCESS CONSTRAINT. In Transmodel it is modeled more extensively and could relate to ACCESS MODE e.g. walk max. 5 minutes, but cycle 10 minutes).</xs:documentation>
 					</xs:annotation>
 				</xs:element>
 			</xs:choice>
-			<xs:element name="IndividualTransportOptions" type="IndividualTransportOptionsStructure" minOccurs="0" maxOccurs="unbounded">
+			<xs:element name="IndividualTransportOptions" type="IndividualTransportOptionStructure" minOccurs="0" maxOccurs="unbounded">
```
trurlurl, line 583:
```suggestion
			<xs:element name="IndividualTransportOption" type="IndividualTransportOptionStructure" minOccurs="0" maxOccurs="unbounded">
```

trurlurl: My idea would have been to do it everywhere on the elements as well.. - or are there too many cases dispersed in all the example files? If yes, and we want do do it, a separate PR would probably be better.

OJP/OJP_JourneySupport.xsd
```diff
@@ -211,48 +211,19 @@
...
 				<xs:annotation>
-					<xs:documentation>Reference to an Operating Day.</xs:documentation>
+					<xs:documentation>MODE OF OPERATION for scheduled and flexible services (NeTEx).</xs:documentation>
 				</xs:annotation>
 			</xs:element>
-			<xs:element ref="siri:VehicleRef" minOccurs="0">
+			<xs:element ref="OperatingDayRef">
 				<xs:annotation>
-					<xs:documentation>Contains the Vehicle reference of the vehicle. In Transmodel this may be the VEHICLE Id.</xs:documentation>
+					<xs:documentation>Reference to an Operating Day.</xs:documentation>
```
skinkie, line 221: Do you want to do OPERATING DAY here?

OJP/OJP_Trips.xsd
```diff
@@ -796,7 +784,10 @@
...
 		</xs:annotation>
 		<xs:sequence>
-			<xs:group ref="StopPointGroup"/>
+			<xs:choice>
+				<xs:group ref="StopPointGroup"/>
+				<xs:element name="GeoPosition" type="siri:LocationStructure"/>
+			</xs:choice>
 			<xs:element name="ServiceArrival" type="ServiceArrivalStructure" minOccurs="0">
 				<xs:annotation>
 					<xs:documentation>describes the arrival situation a this leg board stop point (empty for first leg) ( group of attributes of TIMETABLED PASSING TIME, ESTIMATED PASSING TIME, OBSERVED PASSING TIME)</xs:documentation>
```
skinkie, line 793: Needs rewriting.

ue71603:
Mapping OJP 1.0 modes to OJP 2.0
[MODE_mapping.xlsx](https://github.com/VDVde/OJP/files/12918411/MODE_mapping.xlsx)

trurlurl (dismissed):
!

OJP/OJP_ModesSupport.xsd
```diff
@@ -167,25 +422,73 @@
...
+			<xs:element name="PersonalMode" type="PersonalModesEnumeration" minOccurs="0" maxOccurs="unbounded">
+				<xs:annotation>
+					<xs:documentation>List of personal transport modes to include/exclude. Those are also used in many of the ALTERNATIVE MODES OF OPERATION.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
 			<xs:group ref="siri:PtModeChoiceGroup" minOccurs="0" maxOccurs="unbounded"/>
 		</xs:sequence>
 	</xs:complexType>
-	<xs:complexType name="PrivateModeFilterStructure">
+	<xs:complexType name="ModeAndModeAndOperationFilterStructure">
```
Aurige, line 436: ModeAndMode**And**OperationFilterStructure or ModeAndMode**Of**OperationFilterStructure

OJP/OJP_ModesSupport.xsd
```diff
@@ -1,53 +1,187 @@
...
+				<xs:annotation>
+					<xs:documentation>ALTERNATIVE MODE OF OPERATION. A pooling in relation to park and ride.</xs:documentation>
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="chauffeured">
+				<xs:annotation>
+					<xs:documentation>ALTERNATIVE MODE OF OPERATION for SINGLE JOURNEY with a paid driver. Often more than a single SINGLE JOURNEY.</xs:documentation>
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="privateLift">
```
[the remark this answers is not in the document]
ue71603: asked native English speakers. No response yet. Moved it to PERSONAL MODE OF OPERATION.

OJP/OJP_ModesSupport.xsd
```diff
@@ -1,53 +1,187 @@
...
+				<xs:annotation>
+					<xs:documentation>ALTERNATIVE MODE OF OPERATION demand responsive is used, when there are no SCHEDULED STOP POINTs involved and the timetable component is weak.</xs:documentation>
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="flexibleArea">
+				<xs:annotation>
+					<xs:documentation>Specialisation of the demand responsive ALTERNATIVE MODE OF OPERATION for AREA related offers.</xs:documentation>
+				</xs:annotation>
+			</xs:enumeration>
+			<xs:enumeration value="school">
```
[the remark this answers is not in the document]
ue71603: Moved "school" to ConventionalModeOfOperation. 
Updated comments on enum to clarify which might occur in both CONVENTIONAL and ALTERNATIVE MODE OF OERATION:

OJP/OJP_ModesSupport.xsd
ue71603: fixed

herlitze (approved):

ue71603:
@Aurige  privateLift seems ok to the native English speakers. They also want it in AlternativeMode. What do you think? Move it back there? And then it is usage of a different driver that is the main criterion to have something in Alternative and not in Personal.

OJP/OJP_ModesSupport.xsd
```diff
@@ -167,25 +422,73 @@
...
 		<xs:sequence>
 			<xs:element name="Exclude" type="xs:boolean" default="true" minOccurs="0">
 				<xs:annotation>
-					<xs:documentation>Whether modes in list are to include or exclude from search. Default is exclude.</xs:documentation>
+					<xs:documentation>Whether MODE and MODE OF OPERATION combination in list are to include or exclude from search. Default is exclude.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="PtMode" type="siri:VehicleModesOfTransportEnumeration" minOccurs="0" maxOccurs="unbounded">
+				<xs:annotation>
+					<xs:documentation>List of PT Transport modes to include/exclude. We use the SIRI 2.1 element which allows for many values in the enumeration. In OJP we don't use the following enumeration values: "pti1_xxx", "xxxServices" when a value without the "Service" exists for the same MODE, "allServicesExcept", "selfDrive", "taxi". Also in cases where only a "xxxService" exists in the enumeration, it should be avoided. "other" or "unknown" should only be used, if no other value applies. </xs:documentation>
```
skinkie, line 448:
```suggestion
					<xs:documentation>List of PT Transport modes to include/exclude. We use the SIRI 2.1 element which allows for many values in the enumeration. In OJP we don't use the following enumeration values: "pti1_xxx", "xxxServices" when a value without the "Service" exists for the same MODE, "allServicesExcept", "selfDrive", "taxi". Also in cases where only a "xxxService" exists in the enumeration, it should be avoided. "other" or "unknown" should only be used, if no other value applies.</xs:documentation>
```

OJP/OJP_PlaceSupport.xsd
```diff
@@ -514,6 +514,14 @@
 		<xs:annotation>
 			<xs:documentation>Reference to a geographic PLACE of any type which may be specified as the origin or destination of a trip</xs:documentation>
```
skinkie, line 515:
```suggestion
			<xs:documentation>Reference to a geographic PLACE of any type which may be specified as the origin or destination of a trip.</xs:documentation>
```

OJP/OJP_PlaceSupport.xsd
```diff
@@ -514,6 +514,14 @@
...
 		</xs:annotation>
+		<xs:sequence>
+			<xs:group ref="PlaceRefGroup"/>
+		</xs:sequence>
+	</xs:complexType>
+	<xs:group name="PlaceRefGroup">
+		<xs:annotation>
+			<xs:documentation>Group for reference to a geographic PLACE of any type which may be specified as the origin or destination of a trip</xs:documentation>
```
skinkie, line 523:
```suggestion
			<xs:documentation>Group for reference to a geographic PLACE of any type which may be specified as the origin or destination of a trip.</xs:documentation>
```

skinkie (approved):

herlitze (approved):

trurlurl (approved):

Issue #383, opened 2023-07-19
TripChangeRequest - parameters missing


trurlurl:
I still think that the TripChangeRequest - #381 - needs to know many of the inital parameters that were used to find the trip to be changed. 

Currently, there is only *TripContentFilterGroup* and *OperatorFilter*. 

The easiest would probably be to have the complete *TripParamStructure*, then you can decide which parameters to use and which not. I think the parameter element should be compulsory and the examples have to be adapted.

I might do the PR, but only if @herlitze agrees on the problem, solution, and procedure.

herlitze:
@trurlurl I agree with adding the TripParamStructure, but I think it should be optional as it is optional in the TripRequest as well. I would appreciate it if you could take over the PR, but I can of course correct my carelessness myself.

trurlurl:
Solved by the referring PR.

PR #384, opened 2023-07-19
OJPTripChangeRequest - TripParams added

trurlurl:
This solves #383. The params of the initial TripRequest can now be passed.

I removed two example files since they concerned TripRefineRequest.

@herlitze Can you please check that my understanding of *ChangeLegRef* is correct? I slightly adapted the annotation to clarify the meaning and changed the example to match that understanding.


examples/functions/trip_change/TripChange_001_Request_laterDeparture.xml
```diff
@@ -7,8 +7,13 @@
 			<OJPTripChangeRequest>
 				<siri:RequestTimestamp>2020-06-16T10:00:00Z</siri:RequestTimestamp>
 				<ChangeParams>
-					<ChangeLegRef>3</ChangeLegRef>
+					<ChangeLegRef>2</ChangeLegRef>
 					<EarlierArrivalLaterDeparture>laterDeparture</EarlierArrivalLaterDeparture>
+					<TripParams>
+						<NoStairs>true</NoStairs>
+						<NumberOfResults>3</NumberOfResults>
```
herlitze, line 14: NumberOfResults=3 does not make sense in my opinion, you are just changing one specified trip, so I would expect 1 trip as result.

OJP/OJP_Trips.xsd
```diff
@@ -1768,7 +1768,7 @@
 		<xs:sequence>
 			<xs:element name="ChangeLegRef" type="LegObjectIdType" maxOccurs="unbounded">
 				<xs:annotation>
-					<xs:documentation>Refers to the legs to be changed by the server. </xs:documentation>
+					<xs:documentation>Refers to the transfer legs to be changed by the server. </xs:documentation>
```
herlitze, line 1771: Why would you want to restrict it to transfer legs? There might be future use cases which do not work for transfer legs.
For requesting more transfer time I think both legs could be understood, the transfer leg or the timedLeg before/after. So the change of the examples is fine for me.

herlitze (approved):

OJP/OJP_Trips.xsd
ue71603, line 1771:
```suggestion
					<xs:documentation>Refers to the legs to be changed by the server. </xs:documentation>
```

trurlurl: Hmm, then things are still not completely clear, at least to me.

@herlitze My interpretation was that the *ChangeLegRef* would be the place to specify where exactly in the trip the passenger wants to have more time for an interchange. The *ChangeLegRef* would tell the transfer leg corresponding to this interchange. Another possible interpretation didn't seem to make much sense to me: specifying all the legs that might be changed, because I guess that every leg either before (earlier arrival) or after (later departure) the transfer leg is potentially subject to change.

examples/functions/trip_change/TripChange_001_Request_laterDeparture.xml
ue71603, line 14:
```suggestion
						<NumberOfResults>1</NumberOfResults> <!-- will be ignored as there is one trip that is changed -->
```

OJP/OJP_Trips.xsd
herlitze: @trurlurl Yes, that is my interpretation as well. I am just saying this does not have to be a transfer leg. The TripChange-Request is not used for requesting longer transfers only. When you want to request the times for a specific sharing operator, you would pass the sharing operator you are interested in and the current trip with the generic sharing leg. This sharing leg is not a transfer leg. In the future there might be other use cases as well.

My second argument (For requesting more transfer time I think both legs could be understood, the transfer leg or the timedLeg before/after. So the change of the examples is fine for me.) was for requesting a longer transfer.

1. TimedLeg1
2. TransferLeg
3. TimedLeg2

You have two options you either arrive earlier at the stop of the transfer, or depart later from the stop of the transfer.
Let's say we want to keep TimedLeg1 and want to get a TimedLeg with later departure instead of TimedLeg2. 

To do this you would expect to send ChangeLegRef 2 with laterDeparture, my idea was to send ChangeLegRef3 because it is the leg that actually departs later. In my opinion to send ChangeLegRef 2 the wording of the enum should be changed as well to "Extend to the back/front" instead of later/earlierDeparture.

trurlurl: @herlitze You are right that the wording of the enum should be changed. I did it and adapted the annotations to reflect what you explained. 

Perhaps last question: In the scenario of getting the exact times of two sharing legs, it only seems possible to specify the same operator for both legs?

ue71603 (approved):

herlitze (approved):

skinkie (approved):

Issue #386, opened 2023-07-25
ServiceSections parallel?


ue71603:
Austria wants to use ServiceSections in parallel.
E.g. a person wants to go from Bern to Kerzers. This is part of  coupled journey. So if commerical TrainNumbers are used then there might exist two Service Sections in parallel.

Should we update the description.

Also as the two parallel ServiceSections have different destinations (one going Bern - Neuenburg, the other Bern - Murten), the Origin and Destination should be moved to ServiceSection.

In those cases also Cancellations are on one ServiceSection only. 

Perhaps Section should be merged. 

For possible JSON implementation having groups around everything that can exist n times would be great. Currently ServiceSection is just 1:n

ue71603:
https://github.com/UnionInternationalCheminsdeFer/OSDM/wiki/I-79-OJP-Support-for-parallel-trains

ue71603:
Stefan: Does this involve SIRI.

ue71603:
and also fixing all existing exxamples....

PR #387, opened 2023-07-28
updating Service/ServiceSection

ue71603:
fixing a part of: https://github.com/VDVde/OJP/issues/386

ServiceSection has a lot more functionality now. 

Need to create examples, when generally agreed upon.


skinkie (approved):

OJP/OJP_JourneySupport.xsd
```diff
@@ -59,11 +59,19 @@
...
+			<xs:documentation>Properties of a service combined with a stop sequence interval. In many cases the elements about VEHICLE JOURNEY etc that are bundled in a service in OJP are enough for the whole Leg. However, in some cases enough of these element change within a single leg (usually a PT RIDE LEG), that e.g. the LINE, OPERATOR, FARE SECTION, COMMON SECTION changes. Possible cases are splitting/joining of trains, international trains at the border. Then it becomes necessary to tell at a StopPoint that the information has changed and this results in multiple ServiceSections. This implementation of ServiceSection is a view of some parts of JOURNEY PARTs. Service Sections can exist in parallel (e.g. both coupled JOURNEYS from Bern to Kerzers, one part of the train then goes on to Murten the other to Neuenburg).</xs:documentation>
 		</xs:annotation>
```
herlitze, line 63: @ue71603 what is supposed to happen if the services are only coupled for part of the journey? To stick with your example, when you go from Bern to Neuenburg instead of Kerzers. Should the service to Murten be shown?

OJP/OJP_JourneySupport.xsd
```diff
@@ -88,15 +96,9 @@
 			<xs:documentation>Vehicle journey (not dated).</xs:documentation>
 		</xs:annotation>
 		<xs:sequence>
-			<xs:element ref="JourneyRef"/>
-			<xs:element name="PublicCode" type="xs:normalizedString" minOccurs="0">
-				<xs:annotation>
-					<xs:documentation>Public code of the SERVICE JOURNEY (same meaning as in NeTEx). Examples: "512", "S8" or "Circle Line" or "ICE 488".</xs:documentation>
-				</xs:annotation>
-			</xs:element>
 			<xs:element name="ServiceSection" type="ServiceSectionStructure" minOccurs="1" maxOccurs="unbounded">
```
trurlurl, line 99: I see a circularity - I see *ServiceSectionStructure* containing *StopSequenceIntervalGroup* and *StopSequenceIntervalGroup* containing a *ServiceSectionStructure*.

ue71603: @trurlurl  StopSequenceIntervalGroup does not conatin a ServiceSectionSttructure as far as I can tell.

image

OJP/OJP_JourneySupport.xsd
```diff
@@ -88,15 +96,9 @@
...
 				<xs:annotation>
-					<xs:documentation>Service, can be attributed with a sequence of stop sequence intervals with respective service properties (a view of some parts of JOURNEY PART). The standard case is exactly one Service.</xs:documentation>
+					<xs:documentation>Service, can be attributed with a sequence of stop sequence intervals with respective service properties (a view of some parts of JOURNEY PART). The standard case is exactly one Service. ServiceSection can be in parallel (e.g. in a COUPLED JOURNEY both VEHICLE NUMBERS can be shown when those are commercial.</xs:documentation>
```
skinkie, line 101: You opened a "(" here but did not close it.

ue71603, line 101:
```suggestion
					<xs:documentation>Service, can be attributed with a sequence of stop sequence intervals with respective service properties (a view of some parts of JOURNEY PART). The standard case is exactly one Service. ServiceSection can be in parallel (e.g. in a COUPLED JOURNEY both VEHICLE NUMBERS can be shown when those are commercial).</xs:documentation>
```

OJP/OJP_JourneySupport.xsd
trurlurl: I must have been hallucinating something ...

trurlurl (approved):

skinkie (dismissed):

OJP/OJP_JourneySupport.xsd
herlitze: Yes, it does, thanks for the explanation.

herlitze (commented):
I had a look at TRIAS and the way parallel services are handled there.
I found the way of TimedLeg.ParallelService more intuitive, what was the reason to model it differently?
image

ue71603:
> I had a look at TRIAS and the way parallel services are handled there. I found the way of TimedLeg.ParallelService more intuitive, what was the reason to model it differently? image


Because nobody told me to look there :-)

@skinkie : Should I switch?

skinkie:
I think it is smart to align with Trias. I do hope we don't step on the toes of transmodel.

ue71603:
superseeded with this one: https://github.com/VDVde/OJP/pull/396

Issue #394, opened 2023-08-09
Partial fares


ue71603:
Parrtial fares are already possible with the LegRangeGroup:
![image](https://github.com/VDVde/OJP/assets/24227470/05604614-d446-4d3f-ac2f-f507679616d1)

However, it might not be a range:

E.g.   Walk - Sharing - Transfer - PT1 - Transfer - PT 2 -  Transfer - Demand Responsive - Walk
           1           2              3            4         5            6          7                        8                        9
The PT calculator  can do the PT legs.   In this case:  The range would be 3-7

E.g. Walk - Transfer - PT 1 - Transfer  - PT 2- Walk
         1           2            3           4             5         6
In this case. The leg range group can be omitted, as the whole trip was calculated.

Now to the tricky stuff:
*   Walk   - PT1 - Transfer - Sharing - Transfer - PT 2 -  bicycle
        1         2         3              4             5              6         7

NOVA can here only calculate  1-3 and 5-7 (it even returns probably only 2 and 6)

So what should be the LegRangeGroup here? Or should it return two TripFareResults with the correct LegRange Group. Otherwise it might be necessary to add multiple LegRanges in the LegRangeGroup.

ue71603:
PR done

PR #395, opened 2023-08-15
Improve TripFareResult

ue71603:
Introduce TripId in TripFareResult. Allow for multiple LegRangeGroups in TripFareResult and better explanation of what can be done.

Fixing: https://github.com/VDVde/OJP/issues/394

### DIFF
--- a/OJP/OJP_FareSupport.xsd
+++ b/OJP/OJP_FareSupport.xsd
@@ -697,7 +697,7 @@
 	</xs:group>
 	<xs:complexType name="TripFareResultStructure">
 		<xs:annotation>
-			<xs:documentation>Structure of a Single Trip Fare Request result</xs:documentation>
+			<xs:documentation>Structure of a Single Trip Fare Request result. This may cover the whole trip or only the part the service can answer. If only partial answers can be delivered, then this must be indicated in the error condition (as a warning). For a given trip multiple FareResults may need to be combined to get the whole tariff. A typical example may be classic public transport and a sharing leg. The client must understand that some legs: e.g. TransferLegs and a walking ContinuousLeg will never have a price. The service may include them in one of the tickets. If the whole trip is for free, then a TripFareResult with a price of 0 is expected. It is possible to have combined FareResults (e.g. for a bundle) and in addition price each leg individually.</xs:documentation>
 		</xs:annotation>
 		<xs:sequence>
 			<xs:element name="ErrorCondition" type="OJPErrorStructure" minOccurs="0" maxOccurs="unbounded">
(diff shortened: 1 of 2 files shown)


OJP/OJP_FareSupport.xsd
```diff
@@ -697,15 +697,20 @@
 	</xs:group>
 	<xs:complexType name="TripFareResultStructure">
 		<xs:annotation>
-			<xs:documentation>Structure of a Single Trip Fare Request result</xs:documentation>
+			<xs:documentation>Structure of a single TripFareResult. This may cover the whole trip or only the part the service can answer. If only partial answers can be delivered, then this must be indicated in the error condition (as a warning). For a given trip multiple FareResults may need to be combined to get the whole tariff. A typical example may be classic public transport and a sharing leg. Some legs, e.g. TransferLegs and walking ContinuousLeg will never have a price. The service may include them in one of the tickets. If the whole trip is for free, then a TripFareResult with a price of 0 is expected. It is possible to have combined FareResults (e.g. for a bundle) and in addition price each leg individually.</xs:documentation>
```
sgrossberndt, line 700: Do we need a addidion to the error conditions for the proposed warning?

herlitze (approved):

OJP/OJP_FareSupport.xsd
skinkie: @ue71603 I would agree.

ue71603: `			<xs:enumeration value="FARE_ADDITIONALCHARGES">
				<xs:annotation>
					<xs:documentation>Passengers may be charged additional fees (e.g.: for road tolls or seat reservation).</xs:documentation>
				</xs:annotation>
			</xs:enumeration>
			<xs:enumeration value="FARE_ADDITIONALTICKETS">
				<xs:annotation>
					<xs:documentation>Additional tickets may be necessary because only parts of the passenger trip could be covered.</xs:documentation>
				</xs:annotation>
			</xs:enumeration>`

sgrossberndt: Yes. FARE_ADDITIONALTICKETS should cover it. Maybe it would be good to explicitly name it here then instead of "must be indicated in the error condition (as a warning)"?

ue71603, line 700:
```suggestion
			<xs:documentation>Structure of a single TripFareResult. This may cover the whole trip or only the part the service can answer. If only partial answers can be delivered, then this must be indicated with FARE_ADDITIONALTICKETS (as a warning). For a given trip multiple FareResults may need to be combined to get the whole tariff. A typical example may be classic public transport and a sharing leg. Some legs, e.g. TransferLegs and walking ContinuousLeg will never have a price. The service may include them in one of the tickets. If the whole trip is for free, then a TripFareResult with a price of 0 is expected. It is possible to have combined FareResults (e.g. for a bundle) and in addition price each leg individually.</xs:documentation>
```

trurlurl (approved):

sgrossberndt (approved):
An example covering such a case with a partial result and a `FARE_ADDITIONALTICKETS ` warning would be nice, but not necessary. Otherwise fine for me.

ue71603:
> An example covering such a case with a partial result and a `FARE_ADDITIONALTICKETS ` warning would be nice, but not necessary. Otherwise fine for me.

I suggest to merge it @skinkie  and I will add examples in a separate PR as soon as OJPFareNOVA does those with error messages. Those are then realworld examples: https://github.com/VDVde/OJP/issues/397

skinkie (approved):

Issue #407, opened 2023-10-02
OJPStopEventRequest: Allows PlaceContextStructure (but does not specify behaviour


ue71603:
OJPStopEventRequest allows a full PlaceRef. This means it can be coordinates or an address.

However, it is unclear what the responding system does:
* next stop
* all stops within a radius
* nothing when coordinate/address does not match
Do we need an additional parameter?

How would you describe the behaviour:  @normanoffel  @herlitze 

Affects 7.3.1 in the specification.

PR #412, opened 2023-10-11
clarifying the usage of coord and addresses in the stop event service

ue71603:
fixes https://github.com/VDVde/OJP/issues/407


herlitze (commented):
I disagree with the next stop being always shown. If no stop is reached within range  of limitations , I would not expect a result.

OJP/OJP_StopEvents.xsd
```diff
@@ -15,7 +15,7 @@
 		<xs:sequence>
 			<xs:element name="Location" type="PlaceContextStructure">
 				<xs:annotation>
-					<xs:documentation>LOCATION / PLACE for which to obtain stop event information.</xs:documentation>
+					<xs:documentation>LOCATION / PLACE for which to obtain stop event information. If a coordinate or an address is used, then the result may depend on other parameters: at least the next stop must be shown. All stops in reasonable walking distance and stops that are within range of limitations of the IndividualTransportAptions should be shown.</xs:documentation>
```
ue71603, line 18:
```suggestion
					<xs:documentation>LOCATION / PLACE for which to obtain stop event information. If a coordinate or an address is used, then the result may depend on other parameters: All stops in reasonable walking distance and stops that are within range of limitations of the IndividualTransportAptions should be shown. If the next stop is too far away, then no result is shown. "Reasonable" is usually defined as part of the server configuration.</xs:documentation>
```

ue71603: @herlitze  better?

herlitze (dismissed):

OJP/OJP_StopEvents.xsd
herlitze: Yes, thanks.

skinkie:
@herlitze don't approve if you have open comments ;-)

herlitze:
> @herlitze don't approve if you have open comments ;-)

Oh, you are right Matthias' suggested change is not part of the commit yet.

herlitze (approved):

trurlurl (approved):

skinkie (approved):

PR #419, opened 2023-10-17
updates in the definition of LEG and MODE

ue71603:
From todays meeting. I updated the definitions from C2. I did many things in the specification as well. However, some changes will have to be updated in the specification after #379 is integrated. Sorry, André


OJP/OJP_Trips.xsd
```diff
@@ -575,7 +575,7 @@
 	</xs:simpleType>
 	<xs:complexType name="LegStructure">
 		<xs:annotation>
-			<xs:documentation>a single stage of a TRIP that is made without change of MODE or service (ie: between each interchange)</xs:documentation>
+			<xs:documentation>A single stage of a TRIP that is made without change of MODE or service (ie: between each interchange). Implements LEG from TM 6.</xs:documentation>
```
skinkie, line 578: This is the only place where you mention TM 6, instead of TM 6.2. Is this correct?

ue71603, line 578:
```suggestion
			<xs:documentation>A single stage of a TRIP that is made without change of MODE or service (ie: between each interchange). Implements LEG from TM 6.2.</xs:documentation>
```

OJP/OJP_ModesSupport.xsd
```diff
@@ -351,7 +351,7 @@
 	</xs:annotation>
 	<xs:simpleType name="PersonalModesEnumeration">
 		<xs:annotation>
-			<xs:documentation>TYPE OF MODE than can be used in an individual context.</xs:documentation>
+			<xs:documentation>TYPE OF MODE than can be used in an individual context. PERSONAL MODE in TM 6.2.</xs:documentation>
```
trurlurl, line 354:
```suggestion
			<xs:documentation>TYPE OF MODE that can be used in an individual context. PERSONAL MODE in TM 6.2.</xs:documentation>
```

OJP/OJP_Trips.xsd
```diff
@@ -590,23 +590,23 @@
 			</xs:element>
 			<xs:element name="Duration" type="xs:duration" minOccurs="0">
 				<xs:annotation>
-					<xs:documentation>The duration of the leg (e.g. from Transmodel PT RIDE LEG.Duration)</xs:documentation>
+					<xs:documentation>The duration of the LEG (e.g. from Transmodel PT RIDE LEG.Duration)</xs:documentation>
```
trurlurl, line 593:
```suggestion
					<xs:documentation>The duration of the LEG (e.g. from Transmodel PT RIDE LEG.Duration).</xs:documentation>
```

OJP/OJP_Trips.xsd
ue71603: I think it was a LEG already in TM 6, so I think it is ok.

trurlurl (approved):

skinkie (approved):

PR #420, opened 2023-10-17
privateLift better info

ue71603:


trurlurl:
If ALTERNATIVE MODE OF OPERATION = 
Any publicly advertised mode of operation different from the CONVENTIONAL MODE OF OPERATION, for example: VEHICLE SHARING, VEHICLE RENTAL, VEHICLE POOLING.

and

PERSONAL MODE OF OPERATION = 
A non-advertised mode of operation of vehicles by persons using their own vehicle.

(<https://www.transmodel-cen.eu/model/EARoot/EA10/EA3/EA1/EA3535.htm>)

Then it seems to me that the most clearly distinguishing criterion should be whether it is advertised / commercial or not. PERSONAL = non-advertised, non-commercial

According to Mike's explanation, it is more whether one is on its own or not. PERSONAL = on one's own

Another problem I see: neither of the two criteria justifies to have `hire` in personal and `sharing` in alternative, as they currently are. They are both equally "personal" or unpersonal (personal according to the second criterion, alternative according to the first).

My feeling is that if we keep privateLift in personal, favoring the advertised/commercial criterion, then sharing and hire should be in alternative. 

As for the annotation to privateLift: I wouldn't use the word "stranger", but perhaps: "If the offer is advertised or commercial, then pooling …"

ue71603:
TM 6:
ALTERNATIVE MODE OF OPERATION: Any publicly advertised mode of operation different from the CONVENTIONAL MODE OF OPERATION, for example: VEHICLE SHARING, VEHICLE RENTAL, VEHICLE POOLING.

PERSONAL MODE OF OPERATION: A non-advertised mode of operation of vehicles by persons using their own vehicle.

My feeling: We leave it as it is and let Transmodel work on the MODE OF OPERATION definitions.

@Aurige : What is your take?

ue71603:
@skinkie  @trurlurl  I think we merge this, too. Pls approve, if ok for you @Aurige If you have changes before the merge, then pls do otherwise, we will see what comes out of the Transmodel discussion.

skinkie (dismissed):

trurlurl:
Why not just adhere to the current Transmodel definition, so that things would be clean in this respect? 
That would mean, from my point of view (explained in more detail above): 
* `hire` would have to move to alternative modes;
* The wording "If a stranger ist doing the driving, ..." should be changed to, e.g., "If the offer is advertised or commercial, ..."

ue71603:
@trurlurl  Updated the privateLift description. On hire. I think it really depends on the way you hire. If it is short term andmore in the sense of sharing =>ALTERNATIVE MODE OF OPERATION. If you go on vacation andrent it for two weeks would you really think of this as still an advertised offer?  I have asked Kasia and Christophe and afterwards we can finish this one.

trurlurl (approved):

skinkie (approved):

Aurige (approved):

tim-rivett (approved):

PR #425, opened 2023-10-24
Missing type for IncludeAllRestrictedLines

trurlurl:


trurlurl:
I mistakenly created the branch in a private repository - I suppose it will work anyway? Otherwise I'll try to redo it in the right place.

skinkie:
Not to worry. This is actually how you normally should work, the only down side of this is that we can't make changes and linting is unlikely to work.

skinkie (approved):

ue71603 (approved):

ue71603:
@skinkie Can you pls merge it?

PR #427, opened 2023-10-26
Moving HikingProfile and CyclingProfile out of TripContentFilter

trurlurl:
The proposal is to move HikingProfile and CyclingProfile out of TripContentFilterGroup since they affect the search behavior and not the information to be returned about the trips found (filtering the information would be my interpretation of what TripContentFilter should do).


ue71603 (commented):
The proposed PR removes both Profiles from TripRefineRequest and TripChangeRequest.
Should I add them there as well?

trurlurl:
> The proposed PR removes both Profiles from TripRefineRequest and TripChangeRequest. Should I add them there as well?

TripChangeRequestStructure - TripChangeParam - TripParams = TripParamStructure - TripMobilityFilter contains the profiles.

TripRefineRequest does not. I have a more fundamental question: We only have TripResult of the trip to be refined, this does not seem to include the initial TripParams - how is that sufficient to re-retrieve the trip?

trurlurl:
> > The proposed PR removes both Profiles from TripRefineRequest and TripChangeRequest. Should I add them there as well?
> 
> TripChangeRequestStructure - TripChangeParam - TripParams = TripParamStructure - TripMobilityFilter contains the profiles.
> 
> TripRefineRequest does not. I have a more fundamental question: We only have TripResult of the trip to be refined, this does not seem to include the initial TripParams - how is that sufficient to re-retrieve the trip?

Proposal if it is not sufficient: add new `OriginalTripParams` to `TripRefineParamStructure`.

ue71603 (dismissed):

ue71603:
@herlitze   pls check for the filtergroup I added to trip refinement. I hope this is ok for you as well.

ue71603 (approved):

ue71603:
@skinkie if ok, then pls approve and merge.

skinkie (approved):

PR #439, opened 2024-03-28
Add TripMonitoring

sgrossberndt:
This needs some more work.
- Build is failing due to examples having to be updated to current state
- This branch will be added to develop only after the 2.0 release which will probably have the new xcore documentation rendering. Thus the added [OJP/OJP_TripMonitoring.xsd](https://github.com/VDVde/OJP/compare/develop...feature/trip-monitoring#diff-29a0d05ed34e7039661583357575434d26f4087fd1b84b41912c415b2dfd709e) has to be added to the new documentation rendering. If we do not switch to xcore before merging it it has to be added to the old documentation in schema-collection.xml


trurlurl (dismissed):

sgrossberndt:
This fails due to the examples not having been updated yet.

Issue #442, opened 2024-05-01
Handling of MultiPointType: urgent answer required


ue71603:
Austria noted: 

> Made MultiPointType a required parameter in
> OJPMultiPointTripRequest. #353
> Auf Basis dieser Definition (die jedenfalls Sinn macht, wenngleich die von einem passiven Knoten unterstützten MultiPointTypes ohnehin vorab abgeklärt werden müssen) ist der Parameter MultiPointType in Chapter 8.6.4.5 MultiPointTripParamStructure in Table 151Description of MultiPointTripParamStructure als mandatory angeführt.
> 
> Allerdings sind die Params in denen sich der MultiPointType versteckt laut Chapter 8.6.4.3 OJPMultiPointTripRequestStructure und Table 149 Description of OJPMultiPointTripRequestStructure optional.

The parameter IS optional. and marked that way.
In TripRequest "anyPoint" is the the default value that should be implemented.

In the response it is noted, that it "should" be returned. In my view this might be enough.
Different opinions?

Urgent answers needed.
![image](https://github.com/VDVde/OJP/assets/24227470/f042f48e-86e1-4870-9829-31eb6065198e)

Andrés proposal:

> I agree with the Austrian comment, there is indeed an incoherence and it can be found in the code and the documentation. 
> Correcting it would be simple: setting the cardinality of the Params element in MultiPointTripRequest to 1:1 (instead of 0:1).
> The currently existing inconsistency shows that a certain parameter (MultiPointType) was intended to be a required (non-optional) parameter, but this is not fully implemented / guaranteed by the code.
> P.S. Before applying the change - should we check and discuss whether having it a required parameter is really the right decision - a previous solution (CR #98, still mentioned in the changelog and misleading/superfluous if a required parameter) was to have a default value (anyPoint)? Depending on the decision, I think the changlog entry should be removed.

ue71603:
Stefan/Matthias: Let's also have a default in the response. And update the Changelog (that it is optional with default value)

herlitze:
@ue71603 Do I understand you correctly that you want to keep the OJPMultiPointTripRequest.Params optional and also want to keep OJPMultiPointTripRequest.Params.MultiPointType required? 
The changes you are suggesting are adapting the documentation accordingly and adding a default value to the response (OJPMultiPointTripDelivery.MultiPointType)?
This would be fine for me.

ue71603:
@herlitze  that is exactly what I intend to do...

[...]

PR #443, opened 2024-05-04
Handling of MultiPointType

ue71603:
addresses: https://github.com/VDVde/OJP/issues/442

### DIFF
--- a/OJP/OJP_JourneySupport.xsd
+++ b/OJP/OJP_JourneySupport.xsd
@@ -644,7 +644,7 @@
 			<xs:choice minOccurs="0">
 				<xs:element name="NumberOfResults" type="xs:positiveInteger" minOccurs="0">
 					<xs:annotation>
-						<xs:documentation>The number of trip results that the user wants to see at least (TRIP REQUEST POLICY.maximumNumberofJourneys). Be aware that some implementations will deliver one of the TripResults before the indicated departure time. MultiPointTripRequests may deliver more results than requested depending on the MultiPointType setting. This means one always must check the results against the parameters before using them.</xs:documentation>
+						<xs:documentation>The number of trip results that the user wants to see at least (according to the Transmodel TRIP REQUEST POLICY.maximumNumberofJourneys). Be aware that some implementations will deliver one of the TripResults before the indicated departure time. This means that you can't assume that you get the exact number of results that you indicated in the request from the server.</xs:documentation>
 					</xs:annotation>
 				</xs:element>
 				<xs:group ref="NumberOfResultsGroup" minOccurs="0"/>
--- a/OJP/OJP_JourneySupport.xsd
+++ b/OJP/OJP_JourneySupport.xsd
@@ -644,7 +644,7 @@
 			<xs:choice minOccurs="0">
 				<xs:element name="NumberOfResults" type="xs:positiveInteger" minOccurs="0">
 					<xs:annotation>
-						<xs:documentation>The number of trip results that the user wants to see at least. Be aaware that some implementations will deliver one of the TripResults before the indicated departure time. This means that you can't assume that you get the exact number of results that you asked for in the request from the server.</xs:documentation>
+						<xs:documentation>The number of trip results that the user wants to see at least. Be aware that some implementations will deliver one of the TripResults before the indicated departure time. This means that you can't assume that you get the exact number of results that you asked for in the request from the server.</xs:documentation>
 					</xs:annotation>
 				</xs:element>
 				<xs:group ref="NumberOfResultsGroup" minOccurs="0"/>
--- a/OJP/OJP_JourneySupport.xsd
+++ b/OJP/OJP_JourneySupport.xsd
@@ -644,7 +644,7 @@
 			<xs:choice minOccurs="0">
 				<xs:element name="NumberOfResults" type="xs:positiveInteger" minOccurs="0">
 					<xs:annotation>
-						<xs:documentation>The number of trip results that the user wants to see at least (TRIP REQUEST POLICY.maximumNumberofJourneys). Be aware that some implementations will deliver one of the TripResults before the indicated departure time. This means one always must check the results against the parameters before using them.</xs:documentation>
+						<xs:documentation>The number of trip results that the user wants to see at least (TRIP REQUEST POLICY.maximumNumberofJourneys). Be aware that some implementations will deliver one of the TripResults before the indicated departure time. MultiPointTripRequests may deliver more results than requested depending on the MultiPointType setting. This means one always must check the results against the parameters before using them.</xs:documentation>
 					</xs:annotation>
 				</xs:element>
 				<xs:group ref="NumberOfResultsGroup" minOccurs="0"/>
--- a/OJP/OJP_JourneySupport.xsd
+++ b/OJP/OJP_JourneySupport.xsd
@@ -644,7 +644,7 @@
 			<xs:choice minOccurs="0">
 				<xs:element name="NumberOfResults" type="xs:positiveInteger" minOccurs="0">
 					<xs:annotation>
-						<xs:documentation>The number of trip results that the user wants to see at least (according to the Transmodel TRIP REQUEST POLICY.maximumNumberofJourneys). Be aware that some implementations will deliver one of the TripResults before the indicated departure time. This means that you can't assume that you get the exact number of results that you indicated in the request from the server.</xs:documentation>
+						<xs:documentation>The number of trip results that the user wants to see at least (this is different from what Transmodel suggests with TRIP REQUEST POLICY.maximumNumberofJourneys). Be also aware that some implementations will deliver one of the TripResults before the indicated departure time. This means that you can't assume that you get the exact number of results that you asked for in the request from the server.</xs:documentation>
 					</xs:annotation>
 				</xs:element>
 				<xs:group ref="NumberOfResultsGroup" minOccurs="0"/>
--- a/OJP/OJP_JourneySupport.xsd
+++ b/OJP/OJP_JourneySupport.xsd
@@ -644,7 +644,7 @@
 			<xs:choice minOccurs="0">
 				<xs:element name="NumberOfResults" type="xs:positiveInteger" minOccurs="0">
 					<xs:annotation>
-						<xs:documentation>The number of trip results that the user wants to see at least (this is different from what Transmodel suggests with TRIP REQUEST POLICY.maximumNumberofJourneys). Be also aware that some implementations will deliver one of the TripResults before the indicated departure time. This means that you can't assume that you get the exact number of results that you asked for in the request from the server.</xs:documentation>
+						<xs:documentation>The number of trip results that the user wants to see at least. Be aaware that some implementations will deliver one of the TripResults before the indicated departure time. This means that you can't assume that you get the exact number of results that you asked for in the request from the server.</xs:documentation>
 					</xs:annotation>
 				</xs:element>
 				<xs:group ref="NumberOfResultsGroup" minOccurs="0"/>
--- a/OJP/OJP_Trips.xsd
+++ b/OJP/OJP_Trips.xsd
@@ -1382,7 +1382,7 @@ Can only be used when both participants recognise the same schedule version. If 
 			<xs:group ref="BaseTripPolicyGroup"/>
 			<xs:element name="MultiPointType" type="MultiPointTypeEnumeration" default="anyPoint">
 				<xs:annotation>
-					<xs:documentation>Defines how the router should handle the multiple origins and destinations: If a solution for any one of multiple origin/destination points is sufficient. Or a distinct solution for each of the origin/destination points has to be found. As it is important for the strategy of the distributing trip planning the MultiPointType should be set. If the type is not supported a TRIP_MULTIPOINT_TYPE_NOT_SUPPORTED warning or error must be returned.</xs:documentation>
+					<xs:documentation>Defines how the router should handle the multiple origins and destinations: If a solution for any one of multiple origin/destination points is sufficient. Or a distinct solution for each of the origin/destination points has to be found. As it is important for the strategy of the distributed trip planning the MultiPointType should be set. If the type is not supported a TRIP_MULTIPOINT_TYPE_NOT_SUPPORTED warning or error must be returned. Default is anyPoint.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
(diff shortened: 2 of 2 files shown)


OJP/OJP_Trips.xsd
```diff
@@ -1382,7 +1382,7 @@ Can only be used when both participants recognise the same schedule version. If
 			<xs:group ref="BaseTripPolicyGroup"/>
 			<xs:element name="MultiPointType" type="MultiPointTypeEnumeration" default="anyPoint">
 				<xs:annotation>
-					<xs:documentation>If a solution for any one of multiple origin/destination points is sufficient. Or a distinct solution for each of the origin/destination points has to be found.</xs:documentation>
+					<xs:documentation>Defines how the router should handle the multiple origins and destinations: If a solution for any one of multiple origin/destination points is sufficient. Or a distinct solution for each of the origin/destination points has to be found. As it is important for the strategy of the distributing trip planning the MultiPointType should be set. If the type is not supported a TRIP_MULTIPOINT_TYPE_NOT_SUPPORTED warning or error must be returned.</xs:documentation>
```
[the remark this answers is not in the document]
trurlurl: I added the remark about the default value because that information would otherwise not appear in the documentation and I think readers might still have the same confusion as revealed by the Austrian comment. 

A second point: I myself don't understand the point of keeping *OJPMultiPointTripRequest.Params.MultiPointType* required while *OJPMultiPointTripRequest.Params* is optional. Wouldn't having *MultiPointType* as optional, together with the default *anyPoint* result in *OJPMultiPointTripRequest.Params.MultiPointType*==*anyPoint* in the case if no *OJPMultiPointTripRequest.Params* is provided, as wanted? And wouldn't that avoid the confusion about the change of optional to required in the hierarchy?

herlitze (dismissed):

ue71603:
@skinkie @trurlurl  can ou agree?

trurlurl:
So we now have the parameter as de facto optional and we say that it is important and "should" be set - I think that is OK. But why not more consequently change the cardinality to 0:1 in order to clearly mark it as optional in the documentation table? That would require a slightly greater change in the documentation but it could not lead to confusion any more. (Do we have other places where an element of an optional part is mandatory?)

OJP/OJP_Trips.xsd
```diff
@@ -1382,7 +1382,7 @@ Can only be used when both participants recognise the same schedule version. If
 			<xs:group ref="BaseTripPolicyGroup"/>
 			<xs:element name="MultiPointType" type="MultiPointTypeEnumeration" default="anyPoint">
```
ue71603, line 1383:
```suggestion
			<xs:element name="MultiPointType" type="MultiPointTypeEnumeration" minOccurs="0" default="anyPoint">
```

OJP/OJP_Trips.xsd
```diff
@@ -1430,9 +1427,9 @@ Can only be used when both participants recognise the same schedule version. If
 			<xs:documentation>Multi-point trip response structure.</xs:documentation>
 		</xs:annotation>
 		<xs:sequence>
-			<xs:element name="MultiPointType" type="MultiPointTypeEnumeration" minOccurs="0">
+			<xs:element name="MultiPointType" type="MultiPointTypeEnumeration" minOccurs="0" default="anyPoint">
 				<xs:annotation>
-					<xs:documentation>The MultiPointType should be returned because it may differ from the one asked. Many systems will support only a subset of the MultiPointTypes and it is important to know, what the result is based on.</xs:documentation>
+					<xs:documentation>The MultiPointType used should be returned as it may be different from the one requested. Many systems will only support a subset of the MultiPointTypes, so it is important to know what the result is based on in order to define a strategy for distributed routing.</xs:documentation>
```
sgrossberndt, line 1432: Why is it possible that "MultiPointType" in the response may differ from the requested one?

OJP/OJP_Trips.xsd
```diff
@@ -1412,7 +1409,7 @@ Can only be used when both participants recognise the same schedule version. If
 			</xs:enumeration>
 			<xs:enumeration value="eachOriginDestination">
 				<xs:annotation>
-					<xs:documentation>At least a result for each Origin-Destination pair must be presented.</xs:documentation>
+					<xs:documentation>At least one result for each origin/destination pair must be found.</xs:documentation>
 				</xs:annotation>
 			</xs:enumeration>
 			<xs:enumeration value="somePoints">
```
sgrossberndt, line 1415: > Clarifies that some (probably the "best") origin-destination pairs should be returned.
This documentation should be improved.

OJP/OJP_Trips.xsd
```diff
@@ -1380,24 +1380,21 @@ Can only be used when both participants recognise the same schedule version. If
...
-			<xs:element name="MultiPointType" type="MultiPointTypeEnumeration" default="anyPoint">
+			<xs:element name="MultiPointType" type="MultiPointTypeEnumeration" minOccurs="0" default="anyPoint">
 				<xs:annotation>
-					<xs:documentation>If a solution for any one of multiple origin/destination points is sufficient. Or a distinct solution for each of the origin/destination points has to be found.</xs:documentation>
+					<xs:documentation>Defines how the router should handle requests with multiple origins and destinations. As it is important for the strategy of the distributed trip planning the MultiPointType should be set. If the type is not supported a TRIP_MULTIPOINT_TYPE_NOT_SUPPORTED warning or error must be returned. Default is anyPoint.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
 		</xs:sequence>
 	</xs:group>
 	<xs:simpleType name="MultiPointTypeEnumeration">
```
sgrossberndt, line 1390: What is the difference of `anyPoint` vs `somePoints`?

If I understand it correctly `anyPoint` should return only one trip result, while `somePoints` should return multiple based on some ("probably the best" ?!) selected origin/destination pairs.

But the description of `MultiPointTypeEnumeration` says "`MultiPointType` is more important than `NumberOfResults`. If the algorithm deems it necessary to return more results, it should do so."

So returning multiple results on `anyPoint` is considered valid anyway?

sgrossberndt (changes requested):
I may be quite late to the party regarding the state of OJP 2.0, but the definitions in `MultiPointType` are not really clear to me. Renaming or removing `somePoints` is probably out of question, but we definitely should clarify the documentation. I hope this is still possible?

OJP/OJP_Trips.xsd
ue71603: Because MENTZ and HaCon support only one single MultiPointType. So at least this way we know.

ue71603: ... I think that is exactly what is written there without naming companies.

OJP/OJP_Trips.xsd
ue71603: In anyPoint then it is results from one origin and one destination, but it can still be 10 results all coming from the same pair.
In somePoint one does not know what number of different origins is used and also destinations. It is on up to n.

And yes, having anyPoint with 10 results is valid.

OJP/OJP_Trips.xsd
ue71603: yes.

sgrossberndt (dismissed):

trurlurl (changes requested):

sgrossberndt (approved):

skinkie (dismissed):

OJP/OJP_JourneySupport.xsd
```diff
@@ -644,7 +644,7 @@
 			<xs:choice minOccurs="0">
 				<xs:element name="NumberOfResults" type="xs:positiveInteger" minOccurs="0">
 					<xs:annotation>
-						<xs:documentation>The number of trip results that the user wants to see at least (TRIP REQUEST POLICY.maximumNumberofJourneys). Be aware that some implementations will deliver one of the TripResults before the indicated departure time. This means one always must check the results against the parameters before using them.</xs:documentation>
+						<xs:documentation>The number of trip results that the user wants to see at least (TRIP REQUEST POLICY.maximumNumberofJourneys). Be aware that some implementations will deliver one of the TripResults before the indicated departure time. MultiPointTripRequests may deliver more results than requested depending on the MultiPointType setting. This means one always must check the results against the parameters before using them.</xs:documentation>
```
trurlurl, line 647: What does the *TRIP REQUEST POLICY.maximumNumberofJourneys* refer to? I don't find such a thing in OJP. And it is at odds with "that the user wants to see at *least*".

trurlurl (changes requested):

sgrossberndt:
Please resolve the conflicts so we can get this in.

OJP/OJP_JourneySupport.xsd
```diff
@@ -644,7 +644,7 @@
 			<xs:choice minOccurs="0">
 				<xs:element name="NumberOfResults" type="xs:positiveInteger" minOccurs="0">
 					<xs:annotation>
-						<xs:documentation>The number of trip results that the user wants to see at least (TRIP REQUEST POLICY.maximumNumberofJourneys). Be aware that some implementations will deliver one of the TripResults before the indicated departure time. This means one always must check the results against the parameters before using them.</xs:documentation>
+						<xs:documentation>The number of trip results that the user wants to see at least (according to the Transmodel TRIP REQUEST POLICY.maximumNumberofJourneys). Be aware that some implementations will deliver one of the TripResults before the indicated departure time. This means that you can't assume that you get the exact number of results that you indicated in the request from the server.</xs:documentation>
```
trurlurl, line 647: Yes, this makes it clear where the maximumNumberofJourneys comes from. BUT it still sounds buggy - "least" number of trip results would match *minimumNumberofJourneys* (if this exists), not the current *maximumNumber...*.

ue71603:
@sgrossberndt as soon as André is happy, we can merge this one and then also the release

OJP/OJP_JourneySupport.xsd
trurlurl: @ue71603 @sgrossberndt For me there remains just this as an open point, otherwise everything fine.

sgrossberndt: @trurlurl Did you see the latest correction of @ue71603 ?
https://github.com/VDVde/OJP/pull/443/commits/32b1381f02e0d443a627d86a95cbceac3b61d3e4

sgrossberndt:
> @sgrossberndt as soon as André is happy, we can merge this one and then also the release

@ue71603 What about the [added `minOccurs="0"`](https://github.com/VDVde/OJP/pull/443/files#diff-231d41fa75aa5fbc0757e95a629f845d7143dbb82f7943636c37fd0e3259729bR1383) as discussed [in the issue 442](https://github.com/VDVde/OJP/issues/442#issuecomment-2324678256)

OJP/OJP_JourneySupport.xsd
trurlurl: > @trurlurl Did you see the latest correction of @ue71603 ? [32b1381](https://github.com/VDVde/OJP/commit/32b1381f02e0d443a627d86a95cbceac3b61d3e4)

@sgrossberndt Yes, I've seen the change resolving the initial question. The remaining problem is just that _**maximum**Number_ suggests a "**at most**" while the text says "the number ... the user wants to see **at least**". In Transmodel I've found _MaximumNumberOfResults_, but no corresponding _MinimumNumberOfResults_. I'm forwarding the suggestion of simply removing the reference to Transmodel.

OJP/OJP_JourneySupport.xsd
```diff
@@ -644,7 +644,7 @@
 			<xs:choice minOccurs="0">
 				<xs:element name="NumberOfResults" type="xs:positiveInteger" minOccurs="0">
 					<xs:annotation>
-						<xs:documentation>The number of trip results that the user wants to see at least (TRIP REQUEST POLICY.maximumNumberofJourneys). Be aware that some implementations will deliver one of the TripResults before the indicated departure time. This means one always must check the results against the parameters before using them.</xs:documentation>
+						<xs:documentation>The number of trip results that the user wants to see at least. Be aaware that some implementations will deliver one of the TripResults before the indicated departure time. This means that you can't assume that you get the exact number of results that you asked for in the request from the server.</xs:documentation>
```
trurlurl, line 647:
```suggestion
						<xs:documentation>The number of trip results that the user wants to see at least. Be aware that some implementations will deliver one of the TripResults before the indicated departure time. This means that you can't assume that you get the exact number of results that you asked for in the request from the server.</xs:documentation>
```

trurlurl (approved):

sgrossberndt (changes requested):
Only blocked regarding to minOccurs="0", otherwise I am fine with the change.

OJP/OJP_Trips.xsd
```diff
@@ -1380,44 +1380,41 @@ Can only be used when both participants recognise the same schedule version. If
 		</xs:annotation>
 		<xs:sequence>
 			<xs:group ref="BaseTripPolicyGroup"/>
-			<xs:element name="MultiPointType" type="MultiPointTypeEnumeration" default="anyPoint">
+			<xs:element name="MultiPointType" type="MultiPointTypeEnumeration" minOccurs="0" default="anyPoint">
```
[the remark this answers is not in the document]
trurlurl: CEN is waiting for the documentation text. Can we decide on the final solution? 

I personally would be in favor of the currently proposed solution (cardinality 0:1 and default value), Malte only commented on the previously proposed solution (cardinality 1:1). @herlitze Could you please give your thoughts on the new solution?

herlitze (approved):

ue71603:
@sgrossberndt when ok for you, then we can merge.

OJP/OJP_Trips.xsd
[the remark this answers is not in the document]
sgrossberndt: The default value is `anyPoint` anyway, the cardinality just allows to omit the `MultiPointType` element.
With this change the definition of `MultiPointType` is aligned with the one in the response, so I am fine with it.

sgrossberndt:
I also just found out it had `minOccurs="0"` before until it was removed on 24.05.2023 at 14:43 in "More optimisation methods, reorganising filters and policy for Trip Request according to discussions (#353)".
Since that commit handled multiple things at once it is not clear to me whether this was done on purpose back then.

sgrossberndt:
Apparently it was done on purpose back then by @ue71603:
https://github.com/VDVde/OJP/pull/353#discussion_r1169182424

PR #430, opened 2023-11-07
Switching to 2.0

ue71603:

PR #454, opened 2024-08-24
Update README.md

ue71603:
add new branches


trurlurl (approved):

skinkie (approved):

skinkie:
But I don't think it is correct to say the master branch is set to. So either master is OJP 2. Or we set the default branch to OJP 2. I think the latter is what we want.

skinkie (changes requested):
See suggestion.

sgrossberndt:
I'd like to do a general improvement of the readme before (PR in https://github.com/VDVde/OJP/pull/456) and have a "Release of OJP 2.0" commit. Since the version number has already been set in "Set OJP schema version to 2.0 (#430)" I guess there only the readme to update in order to release 2.0? I guess "Handling of MultiPointType #443" should be merged before? Or is there anything else that needs to be done before?

PR #459, opened 2024-09-09
Params mandatory for MPTR (because of MultiPointType)

ue71603:
based on a comment on the standard by AT.


ue71603:
sorry, just found out, that it is no longer mandatory, but anyPoint is set as the default in #443. So we only have to adapt the specification document . Can you do this André?

PR #463, opened 2024-09-23
Clarify usage of TransferType `changeWithinVehicle`

ue71603:
A possible update of the value. It says this already (at least as an interpretation). Just to clarify.

### DIFF
--- a/OJP/OJP_ModesSupport.xsd
+++ b/OJP/OJP_ModesSupport.xsd
@@ -306,7 +306,7 @@
 			</xs:enumeration>
 			<xs:enumeration value="changeWithinVehicle">
 				<xs:annotation>
-					<xs:documentation>If trains are split, it may be necessary to be in the correct part of the train. Its usage is preferred over remainInVehicle if it is unclear whether the traveller is in the correct part of the train. The possibility of a necessary change within a vehicle leads to the use of changeWithinVehicl</xs:documentation>
+					<xs:documentation>If trains are split, it may be necessary to change to the correct part of the train. Usage of changeWithinVehicle is preferred over remainInVehicle if it is unclear whether the traveller is in the correct part of the train.</xs:documentation>
 				</xs:annotation>
 			</xs:enumeration>
 			<xs:enumeration value="checkIn">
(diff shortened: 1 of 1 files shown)


herlitze (dismissed):

skinkie (dismissed):

OJP/OJP_ModesSupport.xsd
```diff
@@ -306,7 +306,7 @@
 			</xs:enumeration>
 			<xs:enumeration value="changeWithinVehicle">
 				<xs:annotation>
-					<xs:documentation>If namely trains are split, it might be necessary to be in the right part. This is indicated with this value.</xs:documentation>
+					<xs:documentation>If namely trains are split, it might be necessary to be in the right part. This is indicated with this value. This value prerefered to remainInVehicle, when it is unclear, whether the traveler is in the right part of the train. The possibility of a necessary change within a vehicle results in the usage of changeWithinVehicle.</xs:documentation>
```
sgrossberndt: What does "This value `prerefered` to remainInVehicle" mean?

Proposal:

If trains are split, it may be necessary to be in the correct part of the train. Its usage is preferred over remainInVehicle if it is unclear whether the traveller is in the correct part of the train. The possibility of a necessary change within a vehicle leads to the use of changeWithinVehicle.

The sentence "The possibility of a necessary change within a vehicle leads to the use of changeWithinVehicle." is still quite complicated and maybe not even necessary?

sgrossberndt (changes requested):

OJP/OJP_ModesSupport.xsd
ue71603:
```suggestion
					<xs:documentation>If trains are split, it may be necessary to be in the correct part of the train. Its usage is preferred over remainInVehicle if it is unclear whether the traveller is in the correct part of the train. The possibility of a necessary change within a vehicle leads to the use of changeWithinVehicl</xs:documentation>
```

sgrossberndt (approved):

herlitze (approved):

trurlurl (approved):

Issue #467, opened 2025-02-17
Remove unnecessary choice in DatedJourneyStructure


herlitze:

herlitze:
![Image](https://github.com/user-attachments/assets/1aa393bf-d945-4452-9900-40e0ecc94b79)

Choice for a single element does not make sense and confuses code generation tools.

ue71603:
@herlitze is this incorporated in 2.0 already?

mentz-mhe:
@ue71603 no, see https://github.com/VDVde/OJP/pull/469

PR #468, opened 2025-02-17
Remove unnecessary choice in DatedJourneyStructure #467

herlitze:


sgrossberndt:
There are several other places in the OJP code where choices contain only one element. These should be fixed too as well probably? The pull request must be done against branch `develop`, not `main`. Closing this one.

PR #475, opened 2025-08-19
Allow multiple OperatorRefs in AlternativeService

ue71603:
Can be a bug as well. if I should move it to release, pls tell me

Also added the examples and showed how to use it in one.


herlitze (approved):

skinkie (approved):

ue71603:
@skinkie  @Aurige : @herlitze  would prefer this change directly in release 2.0 too. What do you think. It is an oversight on our behalf and nobodys stuff should break. Just add  it there too? Or do we merge the whole develop branch into oit?

PR #494, opened 2026-09-10
Update OJP_Trips.xsd wit ifopt namespace

ue71603:
We seem to have missed sometimes to add the ifopt name space. It often does not pose a problem.


skinkie:
Did we forget it at a point it was used?

Ulf9:
Why is there a IFOPT namespace at all? Is not IFOPT now removed and all of its relevant content now integrated into NeTEx?

ue71603:
@Ulf9  If you want to have it removed, do it in NeTEx first, then Siri  (relevant PR) and then I gladly remove it here. However, I don't start here

PR #495, opened 2026-09-10
added error for problematic projections

### LINKED ISSUE #485: OJPError
For Trips we should have some more error codes (if they are not elsewhere:
* No Result found
* No Trip possible
* Shapes/Detailed path not possible to calculate

ue71603:
fixes #485

we add an error enum saying that the projections/route are problematic


mentz-mhe (commented):
Don't we need the error message for other requests as well?
- TripRefine
- TripChange
- TripInfo
- ...

